PR #8, opened 2020-01-14
core: Logging Middleware with Correlation ID

cstobart:


conf/middleware.py
```diff
@@ -0,0 +1,24 @@
+import logging
+import time
+import uuid
+
+
+class LoggingMiddleware:
```
charlesluokb: class needs to extends `object`?

charlesluokb: if we were working on Python 2, then it is best to extend `object`; in Python 3, class extends `object` explicitly

PR #11, opened 2020-01-15
Adding unit tests

cstobart:


charlesluokb:
@cstobart I've noticed the module serializers.py does not reflect the recent refactoring in models.py, e.g. InvalidEmailSerializer is not required any more, and a new serializer for `LicenceUsage` is needed

PR #12, opened 2020-01-16
Task runner

cstobart:


Pipfile
```diff
@@ -8,7 +8,6 @@ parameterized = "==0.7.0"
 black = "==19.10b0"
 prospector = "==1.1.6.2"
 bandit = "*"
-coverage = ">=4.5.4"
```
charlesluokb: why `coverage` is removed?

docker-compose.yml
```diff
@@ -29,3 +29,40 @@ services:
     expose:
       - 8000
     command: pipenv run ./manage.py runserver 0.0.0.0:8000
+
+  mail:
+    build: ./mailserver
```
charlesluokb: wondering if we should check in the folder ./mailserver for sack of building a full local env?

cstobart: I was thinking of maybe using it as a submodule

Pipfile
cstobart: mistakingly

docker-compose.yml
charlesluokb: I have added the folder `/mailserver`, and updated docker compose file accordingly. If you follow README, then you will have four containers running locally
- lite-hmrc-intg
- lite-hmrc-postgres
- lite-hmrc-ditmail
- lite-hmrc-spiremail

I have also updated .dockerignore to exclude /database and /mailserver when building image, hence we do not need to make /mailserver as a submodule

PR #13, opened 2020-01-16
add folder for building local mail server

charlesluokb:
- add a folder for building a Devcot-based local mail server
- updated docker compose file accordingly


charlesluokb:
@cstobart let me know what you think about this? if you are happy it can be merged into your branch `task-runner`

PR #37, opened 2021-01-29
LTD-585: Parallel run setup and handle licence data extract type

saruniitr:
## Change description
Add new task for parallel run setup. This is to avoid confusion and to ensure we don't send messages to HMRC by mistake.
Add support for handling licence data extract.


mail/urls.py
```diff
@@ -10,4 +10,5 @@
     # path("manage-inbox/", ManageInbox.as_view(), name="manage_inbox"),
     # path("send-usage-updates-to-lite-api/", SendUsageUpdatesToLiteApi.as_view(), name="send_updates_to_lite_api"),
     # path("send-licence-updates-to-hmrc/", SendLicenceUpdatesToHmrc.as_view(), name="send_updates_to_hmrc"),
+    path("parallel-run-outgoing/", TestParallelRunOutgoing.as_view(), name="parallel-run-outgoing"),
```
r4vi, line 13: maybe only enable the test endpoints `if DEBUG`?

saruniitr: :+1:

conf/settings.py
```diff
@@ -103,6 +103,19 @@
 
 DATABASES = {"default": env.db()}
 
+
+INCOMING_EMAIL_PASSWORD = env("INCOMING_EMAIL_PASSWORD")
```
r4vi, line 107: this replaces the `EMAIL_PASSWORD` section?

can we update `local.env` example file?

mail/libraries/mailbox_service.py
```diff
@@ -35,7 +35,7 @@ def read_last_three_emails(pop3connection: POP3_SSL) -> list:
     reversed_mails = list(reversed(mails))
     last_three_mails = reversed_mails[:3]
 
-    message_ids = [get_message_id(line) for line in last_three_mails]
+    message_ids = [get_message_id(line.decode("utf-8")) for line in last_three_mails]
```
r4vi, line 38: this might break tests

r4vi (commented):
it would be good to remove the `parallel_run` name from functions and env vars as much as possible. 

left some small comments

conf/settings.py
r4vi: if you set `local.env` with dummy values it will fix CI

mail/libraries/mailbox_service.py
saruniitr: fixed the tests

r4vi (approved):

PR #41, opened 2021-02-11
LTD-617 Improve code to handle bad emails

r4vi:
This allows the code to continue running and ignore bad email messages that do
not conform to the expected format.

It also adds support for a different mailbox for replies


mail/libraries/data_converters.py
```diff
@@ -48,6 +48,9 @@ def convert_data_for_licence_update(dto: EmailMessageDto) -> dict:
         data["edi_filename"] = dto.attachment[0]
         data["edi_data"] = dto.attachment[1]
 
+    if isinstance(data["edi_data"], bytes):
+        data["edi_data"] = data["edi_data"].decode("utf-8")
```
r4vi, line 52: need to update this to use `DEFAULT_ENCODING`

mail/libraries/mailbox_service.py
```diff
@@ -56,7 +56,21 @@ def mark_status(status):
                 read_status.status = status
                 read_status.save()
 
-            yield to_mail_message_dto(pop3_connection.retr(message_id)), mark_status
+            try:
+                m = pop3_connection.retr(message_id)
+            except error_proto as err:
+                logging.error(f"Unable to RETR message {message_id} in {mailbox_config}: {err}", exc_info=True)
+                mark_status(MailReadStatuses.UNPROCESSABLE)
```
saruniitr, line 63: If there is any network or any other issue while reading that causes an exception and we are not able to read a message, is there a risk that we mark valid messages as unprocessable?

local.env
```diff
@@ -28,6 +28,14 @@ INCOMING_EMAIL_USER=test_user
 INCOMING_EMAIL_POP3_PORT=995
 INCOMING_EMAIL_SMTP_PORT=587
 
+HMRC_TO_DIT_EMAIL_PASSWORD=password
```
saruniitr, line 31: Could you clarify how is this different from the mock hmrc mailbox which is the stand-in for HMRC mailbox?

r4vi: this is not the mock mailbox, this is the return path mailbox. 
Before this change the code was looking at 1 mailbox for both SPIRE-to-HMRC messages and HMRC-to-SPIRE messages.  This splits it up so it looks in both places otherwise it was never seeing replies

mail/libraries/mailbox_service.py
r4vi: Yes that's true but from what I've seen it's usually a malformed message being returned. I did once see this error raised with the server disconnected while trying to RETR.

I can remove the `mark_status` here so it tries them every time?

saruniitr: With retries we may have to limit them again otherwise we end up trying continuously.
If that takes more time may be leave as is and update later? please see whichever is easier.

saruniitr (approved):

mail/libraries/mailbox_service.py
r4vi: I think it's better to err on the side of caution and attempt to process failed ones again - even if it takes a bit longer

PR #45, opened 2021-03-05
LTD-688: Handle licenceReply when building reply messages

saruniitr:
## Change description

The service currently not handling licenceData and licenceReply when
preparing reply messages because of this the run number will comes up
as None both in filename and the content.

Also update serializer to save licenceReply messages.


conf/settings.py
```diff
@@ -133,6 +133,12 @@
 SPIRE_STANDIN_EMAIL_POP3_PORT = env("SPIRE_STANDIN_EMAIL_POP3_PORT")
 SPIRE_STANDIN_EMAIL_SMTP_PORT = env("SPIRE_STANDIN_EMAIL_SMTP_PORT")
 
+HMRC_TO_DIT_EMAIL_PASSWORD = env("HMRC_TO_DIT_EMAIL_PASSWORD")
```
r4vi, line 136: duplicated from 116?

r4vi (changes requested):

conf/settings.py
saruniitr: I think I didn't have latest `master` when I added, will remove this.

r4vi (approved):

PR #49, opened 2021-04-09
LTD-776: Edifact file generation remediation changes

saruniitr:
## Change description

Update edifact file generation as per specification
    
Following things are fixed as per Spec DES236,
 - include reset run number indicator in header
 - Update the licence types as required by spec
 - include country line according to licence type
 - Fix licence line data as per spec
     - pad quantity unit with leading zeros to make it 3 char wide string
    
Update tests

Add validator to ensure file content is as per specification.
Helper functions check each line for expected fields and returns list of errors.

### DIFF
--- a/mail/libraries/builders.py
+++ b/mail/libraries/builders.py
@@ -125,7 +125,7 @@ def build_licence_data_mail(licences) -> Mail:
 
 def build_licence_data_file(licences, run_number) -> (str, str):
     now = timezone.now()
-    file_name = "SPIRE_live_CHIEF_licenceData_{}_{:04d}{:02d}{:02d}{:02d}{:02d}".format(
+    file_name = "SPIRE_LIVE_CHIEF_licenceData_{}_{:04d}{:02d}{:02d}{:02d}{:02d}".format(
         run_number, now.year, now.month, now.day, now.hour, now.minute
     )
 
--- a/mail/libraries/builders.py
+++ b/mail/libraries/builders.py
@@ -125,7 +125,7 @@ def build_licence_data_mail(licences) -> Mail:
 
 def build_licence_data_file(licences, run_number) -> (str, str):
     now = timezone.now()
-    file_name = "SPIRE_LIVE_CHIEF_licenceData_{}_{:04d}{:02d}{:02d}{:02d}{:02d}".format(
+    file_name = "CHIEF_LIVE_SPIRE_licenceData_{}_{:04d}{:02d}{:02d}{:02d}{:02d}".format(
         run_number, now.year, now.month, now.day, now.hour, now.minute
     )
 
--- a/mail/libraries/lite_to_edifact_converter.py
+++ b/mail/libraries/lite_to_edifact_converter.py
@@ -9,7 +9,7 @@ from mail.models import OrganisationIdMapping, GoodIdMapping, LicencePayload
 def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
     now = timezone.now()
     time_stamp = "{:04d}{:02d}{:02d}{:02d}{:02d}".format(now.year, now.month, now.day, now.hour, now.minute)
-    edifact_file = "1\\fileHeader\\SPIRE\\CHIEF\\licenceData\\{}\\{}".format(time_stamp, run_number)
+    edifact_file = "1\\fileHeader\\SPIRE\\CHIEF\\licenceData\\{}\\{}\\Y".format(time_stamp, run_number)
     i = 1
     for licence in licences:
         licence_payload = licence.data
--- a/mail/libraries/lite_to_edifact_converter.py
+++ b/mail/libraries/lite_to_edifact_converter.py
@@ -1,21 +1,16 @@
+import logging
+
 from django.db.models import QuerySet
 from django.utils import timezone
 
-from mail.enums import UnitMapping, LicenceActionEnum, LicenceTypeEnum
+from mail.enums import UnitMapping, LicenceActionEnum, LicenceTypeEnum, LITE_HMRC_LICENCE_TYPE_MAPPING
 from mail.libraries.helpers import get_country_id
 from mail.models import OrganisationIdMapping, GoodIdMapping, LicencePayload
+from mail.libraries.edifact_validator import validate_edifact_file
 
 
-LITE_HMRC_LICENCE_TYPE_MAPPING = {
-    "siel": "SIE",
-    "sicl": "SIE",
-    "sitl": "SIE",
-    "oiel": "OIE",
-    "oiel": "OIE",
-    "ogel": "OGE",
-    "ogcl": "OGE",
-    "ogtl": "OGE",
-}
+class EdifactValidationError(Exception):
+    pass
 
 
 def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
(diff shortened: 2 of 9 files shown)


mail/libraries/builders.py
```diff
@@ -125,7 +125,7 @@ def build_licence_data_mail(licences) -> Mail:
 
 def build_licence_data_file(licences, run_number) -> (str, str):
     now = timezone.now()
-    file_name = "SPIRE_live_CHIEF_licenceData_{}_{:04d}{:02d}{:02d}{:02d}{:02d}".format(
+    file_name = "CHIEF_LIVE_SPIRE_licenceData_{}_{:04d}{:02d}{:02d}{:02d}{:02d}".format(
```
saruniitr, line 128: file name follows the syntax,
<destination>_<integrity>_<source>_<other_fields>...

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -6,49 +6,65 @@
...
+
 def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
     now = timezone.now()
     time_stamp = "{:04d}{:02d}{:02d}{:02d}{:02d}".format(now.year, now.month, now.day, now.hour, now.minute)
-    edifact_file = "1\\fileHeader\\SPIRE\\CHIEF\\licenceData\\{}\\{}".format(time_stamp, run_number)
-    i = 1
+
+    edifact_file = "1\\fileHeader\\SPIRE\\CHIEF\\licenceData\\{}\\{}\\Y".format(time_stamp, run_number)
+
+    line_no = 1
```
saruniitr, line 27: just a meaningful name instead of `i`

mail/libraries/builders.py
```diff
@@ -111,6 +111,9 @@ def build_licence_data_mail(licences) -> Mail:
     last_lite_update = LicenceData.objects.last()
     run_number = last_lite_update.hmrc_run_number + 1 if last_lite_update else 1
     file_name, file_content = build_licence_data_file(licences, run_number)
+    if file_content is None:
+        return None
```
alixedi, line 115: I am curious to learn why we are having to deal with `file_content` being `None` at this point?

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -96,34 +111,40 @@ def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
...
+    line_no += 1
     edifact_file += "\n{}\\fileTrailer\\{}".format(
-        i, licences.count() + licences.filter(action=LicenceActionEnum.UPDATE).count()
+        line_no, licences.count() + licences.filter(action=LicenceActionEnum.UPDATE).count()
     )
+
+    errors = validate_edifact_file(edifact_file)
+    if errors:
+        logging.error(f"File content not as per specification, {errors}")
+        return None
```
alixedi, line 146: Might be better to raise an Exception here instead of logging manually and returning with `None`.

I think (although I am not sure) that this would take care of the -

```
if file_content is None:
         return None
```

Earlier in this PR.

mail/tasks.py
```diff
@@ -211,6 +211,10 @@ def send_licence_data_to_hmrc():
                 return
 
             mail = build_licence_data_mail(licences)
+            if mail is None:
+                logging.error("Mail not created, errors in edifact file")
+                return
```
alixedi, line 216: Same as the earlier comment. 

It is more Pythonic to raise and catch specific exceptions instead of relying on the return type to be `None`.

saruniitr: agree, updating this.

mail/libraries/lite_to_edifact_converter.py
saruniitr: :+1:

mail/libraries/builders.py
saruniitr: `file_content` here is the edifact_file, currently I am returning `None` if there is any validation failure but yes raising exception makes sense as we cannot continue further with this failure.

mail/tasks.py
```diff
@@ -219,6 +220,8 @@ def send_licence_data_to_hmrc():
             send(server, mail_dto)
             update_mail(mail, mail_dto)
             licences.update(is_processed=True)
+    except EdifactValidationError as err:  # noqa
```
saruniitr, line 223: can be combined with below `Exception` if that is preferred

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -6,49 +6,65 @@
...
+                "E",  # Export use
                 old_payload.get("start_date").replace("-", ""),
                 old_payload.get("end_date").replace("-", ""),
             )
-            i += 1
-            edifact_file += "\n{}\\end\\licence\\{}".format(i, i - start_line)
-            start_line = i
-            i += 1
+            line_no += 1
+            edifact_file += "\n{}\\end\\licence\\{}".format(line_no, line_no - start_line)
```
alixedi, line 43: This is a very good opportunity to replace `format` with f-strings.

alixedi, line 43: Might be worth having an `edifact_file_lines` list instead of an `edifact_lines` string.

Append new lines to the list, `return edifact_file_lines.join("\n")` at the end.

alixedi, line 43: One more comment on this. 

Perhaps even better than f-strings (or format) is -

```
[line_no, "license", get_transaction_refere...].join('\\')
```

It is more DRY and might expose some refactoring opportunities e.g. functions like `license_line(line_no...)`, `trader_line`, `country_line` etc. which will allow us to write better unit tests.

This might be too big for this PR though so happy dor you to `join("\\")` for now and we will circle back later and fix this.

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -74,19 +90,27 @@ def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
...
+            if licence_payload.get("type") in LicenceTypeEnum.OPEN_LICENCES:
+                if licence_payload.get("country_group"):
+                    line_no += 1
+                    edifact_file += "\n{}\\country\\\\{}\\{}".format(line_no, licence_payload.get("country_group"), "D")
+                elif licence_payload.get("countries"):
+                    for country in licence_payload.get("countries"):
+                        country_id = get_country_id(country)
+                        line_no += 1
+                        edifact_file += "\n{}\\country\\{}\\\\{}".format(line_no, country_id, "D")
+            elif licence_payload.get("type") in LicenceTypeEnum.STANDARD_LICENCES:
```
alixedi, line 97: Using `elif` here comes across as a little fishy because we don't have an `else` clause?

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -139,15 +134,21 @@ def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
...
         line_no += 1
         edifact_file += "\n{}\\end\\licence\\{}".format(line_no, line_no - start_line)
     line_no += 1
     edifact_file += "\n{}\\fileTrailer\\{}".format(
         line_no, licences.count() + licences.filter(action=LicenceActionEnum.UPDATE).count()
     )
+
+    errors = validate_edifact_file(edifact_file)
+    if errors:
+        logging.error(f"File content not as per specification, {errors}")
```
alixedi, line 149: 1. You don't need to return errors from `validate_edifact_file`. You can raise the exception directly from that function. `errors` could be a property on that exception and you even write a `__repr__` function that does something like `'\n'.join(errors)`.

2. You don't need to log the error here. If you just raise it (without catching it), it will appear in the logs.

mail/libraries/lite_to_edifact_converter.py
alixedi: Even better (I might be wrong) to have an `edifact_file` list of lists and -

```
return '\n'.join(['\\'.join(line) for line in edifact_file])
```

Again, this is DRY and probably worth doing because it encodes the format in a way that decreases the possibility of mistakes.

mail/libraries/lite_to_edifact_converter.py
saruniitr: True but my idea is to keep the validator code not coupled because it is at very initial state and testing is easier if they don't have to deal with exceptions, that is why for now it is just returning a list of errors and caller can take the appropriate action giving us some flexibility.

alixedi (approved):
LGTM.

PR #50, opened 2021-04-19
LTD-787: Edifact address line fixes

saruniitr:
## Change description

The length of each address line has a limit as per specification. This PR updates the generation code to validate for these limits and report errors accordingly.

The foreign trader address is currently captured as a single text field and we cannot control the format in which the user provides this address. To stick to the limits as per spec this address is split into chunks and the file is generated accordingly.

Updated tests and added new ones to test the new functionality.

### DIFF
--- a/mail/libraries/edifact_validator.py
+++ b/mail/libraries/edifact_validator.py
@@ -1,8 +1,12 @@
+import re
+
 from mail.enums import LicenceActionEnum, LITE_HMRC_LICENCE_TYPE_MAPPING
 
 FILE_HEADER_FIELDS_LEN = 8
 LICENCE_TRANSACTION_HEADER_FIELDS_LEN = 9
 PERMITTED_TRADER_HEADER_FIELDS_LEN = 13
+PERMITTED_TRADER_NAME_MAX_LEN = 80
+PERMITTED_TRADER_ADDR_LINE_MAX_LEN = 35
 COUNTRY_FIELDS_LEN = 5
 FOREIGN_TRADER_FIELDS_LEN = 10
 LICENCE_LINE_FIELDS_LEN = 13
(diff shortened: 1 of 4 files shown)


mail/libraries/edifact_validator.py
```diff
@@ -66,6 +75,22 @@ def validate_licence_transaction_header(data_identifier, record):
...
+    space_regex = re.compile(r" *(%s)$" % incode_pattern)
+
+    postcode = value.upper().strip()
+    # Put a single space before the incode (second part).
+    postcode = space_regex.sub(r" \1", postcode)
+
+    if not postcode_regex.search(postcode):
+        return False
+
+    return True
```
alixedi, line 91: Is this regex for *all* addresses anywhere in the world? Can't understand how it is possible. Might be a good idea if you can share the source of it?

alixedi: Another point I would like to make here (although it's a broader point) is how are we planning to recover from e.g. an invalid postcode?

Generally, validation for data is handled as it is being ingested into a system.

saruniitr: [1] This is mainly for UK based postcodes as permitted trader should be a UK address for Exports. For foreign trader (eg End user) it will be a foreign address and no validation on the postcode is done. In this postcode is also optional.

This was existing code and I have taken it from the user registration code [(validate_postcode())](https://github.com/uktrade/lite-api/blob/master/api/addresses/serializers.py#L25)

[2] Regarding recovery - Since we first encounter this during User registration we report this to the user if there is invalid postcode, unless this is corrected the registration cannot be completed successfully. By the time we actually use it in `lite-hmrc` it is already validated and has correct postcode but I am checking it here again for completeness.
Previously it failed I think it is because that organisation was created before any of this validation is in place.

alixedi: Cool! Agree on both points. Thank you. For reference, would be nice to include a little comment that the validation logic was copied from `lite.api.<something>.validate_postcode`.

alixedi (approved):

PR #52, opened 2021-04-28
LTD-787: Fix TraderId and licence line issues in Edifact file generation

saruniitr:
## Change description

Testing showed that we are not including a valid TraderId and that the licence line missing some of the optional fields.
Adds changes to include the TraderId (EORI number) and missing optional fields. These are optional because they depend on whether the product is controlled by quantity or value but we need to include them as null even if they are not applicable otherwise they fail HMRC validation and they cannot generate usage data.


r4vi (approved):
the backslash escaping is blowing my mind

saruniitr:
The spec says some of these can be absent if controlled by Quantity - but not clear if absent meaning totally remove them or can be left empty. Because the files are accepted without these earlier but seeing issues when going for usage details. I guess it is safe to include them as empty anyway.

r4vi:
> I guess it is safe to include them as empty anyway. 
Yes since it fails when they generate usage if they're not there, we should add them

PR #54, opened 2021-05-10
LTD-787: Usage data processing and outstanding issues fixes

saruniitr:
## Change description

Add changes for the issues encountered while processing the usage data messages from HMRC.

1. In preparation to retrieve `Message-ID` from the message and to save in `MailReadStatus` model, existing `message_id` is renamed to `message_num` as we currently save message number instead of id.
2. Add new migration to save the `message_id`
3. Update email retrieval code to inspect the header, retrieve `Message-ID` and save it in the model. Previously we were using message number which is an `int` and it is observed that they are getting re-used. Because of this we sometimes ignore new messages as READ/PROCESSED. Now we use `Message-ID` which is unique for each message.
4. Update mail status to `reply_sent` in case of usage_data because in this case we are not expecting usage reply either from SPIRE or LITE and we are also not sending usage replies to HMRC. If it is left in pending state we won't be able to process any other messages hence handle this special case and update the status to `reply_sent`.


mail/models.py
```diff
@@ -175,17 +175,17 @@ class Meta:
 
 class MailboxConfig(TimeStampedModel):
     username = models.TextField(null=False, blank=False, primary_key=True, help_text="Username of the POP3 mailbox")
-    start_message_id = models.TextField(
-        null=True,
-        blank=False,
-        default=None,
-        help_text="Process messages from this message ID onwards, can be Null if messages should be processed from the very first message in the mailbox",
-    )
```
alixedi, line 183: This seems like some kind of a pointer on a queue. 

Are we not using this anymore?

mail/libraries/mailbox_service.py
```diff
@@ -15,23 +15,38 @@ def send_email(smtp_connection: SMTP, message: Message):
     smtp_connection.send_message(message)
 
 
-def get_message_id(listing_msg: bytes) -> bytes:
+def get_message_id(pop3_connection, listing_msg):
```
alixedi, line 18: Would it be useful if we remove `listing_msg` from here and instead use `pop3_connection.list()` first thing in the function?

mail/libraries/mailbox_service.py
```diff
@@ -40,10 +55,10 @@ def get_message_iterator(pop3_connection: POP3_SSL, username: str) -> Iterator[T
         ).values_list("message_id", flat=True)
     )
 
-    for message_id in mail_message_ids:
+    for message_id, message_num in mail_message_ids:
         # only return messages we haven't seen before
         if message_id not in read_messages:
```
alixedi, line 60: If we use `create` instead of `get_or_create` we may be able to remove this check as well as - 

https://github.com/uktrade/lite-hmrc/pull/54/commits/237613730245a3c811f087e864c5143c7de9fe84#diff-7cccd0c25c13c4fdaed21bb6999585a73109a833b1980453fad2ca8e51901db7R52-R56

(By catching the "this already exists" exception)

mail/libraries/mailbox_service.py
alixedi: (this will be a useful refactor but please ignore if you are pressed for time)

mail/libraries/mailbox_service.py
```diff
@@ -70,8 +85,8 @@ def mark_status(status):
 
 def read_last_message(pop3_connection: POP3_SSL) -> EmailMessageDto:
     _, mails, _ = pop3_connection.list()
-    message_id = get_message_id(mails[-1])
-    return to_mail_message_dto(pop3_connection.retr(message_id))
+    _, message_num = get_message_id(pop3_connection, mails[-1])
+    return to_mail_message_dto(pop3_connection.retr(message_num))
```
alixedi, line 89: Should we not catch the exception here like we do here?

https://github.com/uktrade/lite-hmrc/pull/54/files/744bfcf2f5ffec69cd9b4a43fc84f4ec12fb1bb8..237613730245a3c811f087e864c5143c7de9fe84#diff-7cccd0c25c13c4fdaed21bb6999585a73109a833b1980453fad2ca8e51901db7R71

mail/libraries/mailbox_service.py
saruniitr: ok so pull the list within this function and return list of `(Message-Id, message_num)` - agree this is better. I will update this.

mail/libraries/mailbox_service.py
saruniitr: Yes we do, I think it was missed out when we first added this.

mail/models.py
saruniitr: It is a pointer but even in the current code we are not using this. This was introduced early on to start from a particular message number but now we check the status READ/UNPROCESSABLE etc and skip them.

Going forward it will not be of any use because we are going to use `Message-ID` taken from the message header which is a uuid and not sequential hence I removed this.

mail/libraries/mailbox_service.py
saruniitr: I thought of changing this, on a second thought I would like to keep as is mainly because I do not want to add another `try ... catch` here as we already have two below. It will hurt readability.

mail/libraries/mailbox_service.py
saruniitr: Modified this to pull the list of messages but it looks like this is meant to be used with a single message. This function is used in three different places in this file. 1. It goes through all messages, 2. just last three messages, 3. just the last message.

We can refactor and make it work by passing in the number of messages, eg,
1 - read last message
3 - read last three messages
None - as default arg to pull all messages.
In all these cases we have to return the list of tuples `[(message_id, message_num) ...]` which calls for some other changes on the caller side.

Just want to gather your thoughts before I proceed to change like this.

alixedi: Ty for looking into this. 

Given that this is used in 3 different places, maybe it's best to leave it for after the launch.

alixedi (approved):

mail/libraries/mailbox_service.py
saruniitr: yes, will leave as is.

PR #60, opened 2021-05-18
LTD-787: Remove content transfer encoding of 7-bit from message header

saruniitr:
HMRC not able to process the file so this is a potential issue identified.


alixedi (approved):

mail/models.py
```diff
@@ -183,8 +183,8 @@ class MailReadStatus(TimeStampedModel):
         default="",
         help_text="Sequence number of the message as assigned by pop3 when the messages list is requested from the mailbox",
     )
-    message_id = models.UUIDField(
-        default=uuid.uuid4, help_text="Unique Message-ID of the message that is retrieved from the message header"
+    message_id = models.TextField(
+        default="", help_text="Unique Message-ID of the message that is retrieved from the message header"
```
alixedi, line 187: Might want to keep the default Arun.

mail/models.py
```diff
@@ -183,8 +183,8 @@ class MailReadStatus(TimeStampedModel):
         default="",
         help_text="Sequence number of the message as assigned by pop3 when the messages list is requested from the mailbox",
     )
-    message_id = models.UUIDField(
-        default=uuid.uuid4, help_text="Unique Message-ID of the message that is retrieved from the message header"
+    message_id = models.TextField(
```
agalamatis, line 186: why TextField instead of CharField?

alixedi (approved):
One comment that I will leave with you.

mail/models.py
saruniitr: we need to define max_length if it is CharField, Textfield is actually and no difference internally anyway

alixedi: Prolly didn't want to commit to specifying `max_length`.

saruniitr: Yeah I don't know the `max_length`, there is no performance hit also.

conf/settings.py
```diff
@@ -223,6 +224,7 @@
 BACKGROUND_TASK_RUN_ASYNC = True
 # Number of times a task is retried given a failure occurs with exponential back-off = ((current_attempt ** 4) + 5)
 MAX_ATTEMPTS = 7  # e.g. 7th attempt occurs approx 40 minutes after 1st attempt (assuming instantaneous failures)
+BACKGROUND_TASK_MAX_RUN_TIME = env("BACKGROUND_TASK_MAX_RUN_TIME", default=900)
```
alixedi, line 227: @saruniitr This may not work because of [https://github.com/arteria/django-background-tasks/blob/master/background_task/settings.py#L30-L33](this).

Might want to use the documented - `MAX_RUN_TIME`.

alixedi: While you are at it, what do you think about - https://github.com/arteria/django-background-tasks/blob/master/background_task/management/commands/process_tasks.py#L55-L59

IDK exactly what it does.

PR #65, opened 2021-05-25
LTD-787: Sort the message dto based on date before serializing them

saruniitr:
This ensure we process the emails in the same order as they are received


mail/libraries/routing_controller.py
```diff
@@ -78,13 +78,19 @@ def check_and_route_emails():
...
 
     spire_to_dit_server = get_spire_to_dit_mailserver()
     if hmrc_to_dit_server != spire_to_dit_server:
         # if the config for the return path is different to outgoing mail path
         # then check the return path otherwise don't bother as it will contain the
         # same emails.
-        email_message_dtos.extend(_get_email_message_dtos(spire_to_dit_server))
+        reply_message_dtos = _get_email_message_dtos(spire_to_dit_server)
+        reply_message_dtos = sort_dtos_by_date(reply_message_dtos)
+        logging.info(f"Reply message dtos sorted by date: {email_message_dtos}")
```
alixedi, line 91: `Black` should tell you to use `%` for logging and not f-strings b/c lazy evaluation.

alixedi (approved):

mail/libraries/routing_controller.py
saruniitr: No it didn't complain, only formatting errors

PR #69, opened 2021-05-26
Check only the last 500 emails when looking for new emails in the mail…

saruniitr:
Since we don't delete emails from these mailboxes the total number can be
very high over a period which increases the processing time.

The list we pull from pop3 gives us message number and size eg b"552 42569".
As we don't delete emails the message number is an increasing value so the
new emails will always be at the end and checking last emails should be sufficient.


mail/libraries/mailbox_service.py
```diff
@@ -73,7 +73,12 @@ def get_message_iterator(pop3_connection: POP3_SSL, username: str) -> Iterator[T
     _, mails, _ = pop3_connection.list()
     mailbox_config, _ = MailboxConfig.objects.get_or_create(username=username)
 
-    mail_message_ids = [get_message_id(pop3_connection, m.decode(settings.DEFAULT_ENCODING)) for m in mails]
+    # Check only the last 50 emails
```
alixedi, line 76: Any particular reason we are picking the last 50 emails?

saruniitr: no specific reason, I have just picked atleast two days count.

alixedi: How can we assume that 50 emails are at least two days count?

saruniitr: Well by looking at the past data, it can be made 75 or 100 also as it doesn't make much difference as long it is not 1000s. What number would you recommend?

alixedi: I think we agreed on 500.

saruniitr: yes, changed to 500. Initially forgot to update the comment.

alixedi (approved):

PR #75, opened 2021-07-28
Ltd 979 license status query

dgnball:
Add an endpoint to lite-hmrc that takes a license id and returns back the mail status.

### DIFF
--- a/mail/tests/test_end_to_end.py
+++ b/mail/tests/test_end_to_end.py
@@ -1,3 +1,5 @@
+from urllib.parse import quote
+
 from conf.settings import MAILHOG_URL
 
 import requests
@@ -40,3 +42,6 @@ class EndToEndTests(LiteHMRCTestClient):
 8\end\licence\7
 9\fileTrailer\1"""
         assert body == expected_mail_body  # nosec
+        encoded_reference_code = quote("GBSIEL/2020/0000001/P", safe='')
+        license = self.client.get(f"/mail/license/?id={encoded_reference_code}")
+        assert "reply_pending" == license["status"]
--- a/mail/tests/test_end_to_end.py
+++ b/mail/tests/test_end_to_end.py
@@ -44,4 +44,4 @@ class EndToEndTests(LiteHMRCTestClient):
         assert body == expected_mail_body  # nosec
         encoded_reference_code = quote("GBSIEL/2020/0000001/P", safe="")
         response = self.client.get(f"/mail/license/?id={encoded_reference_code}")
-        assert "reply_pending" == response.json()["status"]
+        assert "reply_pending" == response.json()["status"]  # nosec
--- a/mail/tests/test_end_to_end.py
+++ b/mail/tests/test_end_to_end.py
@@ -42,6 +42,6 @@ class EndToEndTests(LiteHMRCTestClient):
 8\end\licence\7
 9\fileTrailer\1"""
         assert body == expected_mail_body  # nosec
-        encoded_reference_code = quote("GBSIEL/2020/0000001/P", safe='')
-        license = self.client.get(f"/mail/license/?id={encoded_reference_code}")
-        assert "reply_pending" == license["status"]
+        encoded_reference_code = quote("GBSIEL/2020/0000001/P", safe="")
+        response = self.client.get(f"/mail/license/?id={encoded_reference_code}")
+        assert "reply_pending" == response.json()["status"]
--- a/mail/tests/test_end_to_end.py
+++ b/mail/tests/test_end_to_end.py
@@ -43,5 +43,5 @@ class EndToEndTests(LiteHMRCTestClient):
 9\fileTrailer\1"""
         assert body == expected_mail_body  # nosec
         encoded_reference_code = quote("GBSIEL/2020/0000001/P", safe="")
-        response = self.client.get(f"/mail/license/?id={encoded_reference_code}")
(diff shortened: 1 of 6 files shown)


mail/views.py
```diff
@@ -111,3 +110,10 @@ class SetAllToReplySent(APIView):
     def get(self, _):
         Mail.objects.all().update(status=ReceptionStatusEnum.REPLY_SENT)
         return HttpResponse(status=HTTP_200_OK)
+
+
+class License(APIView):
+    def get(self, request):
+        license_id = request.GET.get("id", "")
+        mail = list(LicenceData.objects.filter(licence_ids__contains=license_id))[0].mail
+        return JsonResponse({"status": mail.status})
```
alixedi, line 119: I would recommend writing a model serializer if we don't already have one and [using DRF to do the work for you](https://www.django-rest-framework.org/api-guide/generic-views/#examples).

mail/tests/test_end_to_end.py
```diff
@@ -40,3 +42,6 @@ def test_send_email_to_hmrc_e2e(self):
 8\end\licence\7
 9\fileTrailer\1"""
         assert body == expected_mail_body  # nosec
+        encoded_reference_code = quote("GBSIEL/2020/0000001/P", safe="")
+        response = self.client.get(f"/mail/license/?id={encoded_reference_code}")
+        assert "reply_pending" == response.json()["status"]  # nosec
```
alixedi, line 47: Generally, we use the `reverse` method that comes with Django to work out the URL. You don't even need to call the `quote` if you use it.

Also, a small one but I think generally the standard way to write assertions is -

```
assert response.json()["status"] == "reply_pending"
```

(It is a check whether reality is the same as expectations and so it reads better in that order)

mail/views.py
```diff
@@ -111,3 +110,10 @@ class SetAllToReplySent(APIView):
     def get(self, _):
         Mail.objects.all().update(status=ReceptionStatusEnum.REPLY_SENT)
         return HttpResponse(status=HTTP_200_OK)
+
+
+class License(APIView):
+    def get(self, request):
+        license_id = request.GET.get("id", "")
+        mail = list(LicenceData.objects.filter(licence_ids__contains=license_id))[0].mail
```
eadpearce, line 118: if it's only expected to ever return 1 object you can use .get() instead

mail/views.py
```diff
@@ -111,3 +110,10 @@ class SetAllToReplySent(APIView):
     def get(self, _):
         Mail.objects.all().update(status=ReceptionStatusEnum.REPLY_SENT)
         return HttpResponse(status=HTTP_200_OK)
+
+
+class License(APIView):
```
eadpearce, line 115: if we're just returning the status of the mail a better name would be LicenseMailStatus

alixedi (commented):
We like to organize our commits into small atomic changes that read well in `git blame` e.g. this work can be squashed into a single commit - "Add endpoint to returns the list of licenses."

mail/views.py
dgnball: Don't we get diminishing returns from using a serializer if we are only pulling out a single field?

mail/views.py
dgnball: Good spot! Will do.

mail/views.py
dgnball: I see your point and I did conisder something like this but the URL is already /mail/license and I don't know if we want more mail-license related info up the pipe in the future so I'm going to leave the name unless people feel very strongly about it

mail/tests/test_end_to_end.py
dgnball: Happy to swap the assert round. Not that happy about using reverse. I prefer to see the URL in it's natural state as it provides free documentation of how to use the endpoint. I don't really see what benefit reverse gives us.

dgnball:
> We like to organize our commits into small atomic changes that read well in `git blame` e.g. this work can be squashed into a single commit - "Add endpoint to returns licenses."

ok, will do

mail/tests/test_end_to_end.py
saruniitr: `reverse` is better. If we change the path eg `/mail/licence/` to `/main/something-else/etc/etc` then we don't have to change here, it is a best practice.

mail/views.py
saruniitr: We don't fully know what details we want to export using this endpoint.
I agree `serializer` is better then the same can be used in tests.

mail/tests/test_end_to_end.py
dgnball: I don't think you can use reverse with query strings. I tried and it didn't work for me.

dgnball: See https://gist.github.com/benbacardi/227f924ec1d9bedd242b

mail/views.py
dgnball: Ok, done.

mail/tests/test_end_to_end.py
alixedi: Good catch! I reckon we should still use `reverse` but use f-strings to pop query string params into the URL.

dgnball: Ok, done!

saruniitr (approved):
just spelling change required for License.
Every where it is `Licence` but here it is `License`.

Once that is correct it can be merged.

PR #76, opened 2021-07-29
LTD-1056: Remove unique_together constraint on TransactionMapping model

saruniitr:
## Change description

This model is used to record the usage data transactions for a given licence.
For reasons unknown there is a unique_together constraint enforced on this based
on certain fields even though we are not using this model for any decision
making etc.

If usage is recorded for any good of a licence then the usage data file coming
from hmrc is containing multiple licenceUsage lines in the file and because of
the unique_together constraint we were not able to separate LITE specific
usage data and record it correctly in our system.

Fixing this by removing the constraint, this should allow us to record the
usage for LITE licences correctly.

Please note this won't affect any of the SPIRE licences.


mail/models.py
```diff
@@ -166,9 +166,6 @@ class TransactionMapping(models.Model):
     usage_transaction = models.CharField(null=False, blank=False, max_length=35)
     usage_data = models.ForeignKey(UsageData, on_delete=models.DO_NOTHING)
 
-    class Meta:
-        unique_together = [["licence_reference", "line_number", "usage_data"]]
-
```
saruniitr, line 171: this was preventing from sending the usage data correctly to LITE. During the separation of LITE licences usage data it was trying to create instances of `TransactionMapping` and failing because of this constraint.
It is safe to remove this as these are not used for any decision making, it is also not clear why it is constrained in the first place.

agalamatis (approved):
Try it on staging ;)

PR #78, opened 2021-08-04
LTD-1088: Expose Mail queue status to Sentry periodically

saruniitr:
## Change description

During every polling session export the queue status to Sentry so that we
can check the status without logging into the mailboxes.

### DIFF
--- a/mail/libraries/helpers.py
+++ b/mail/libraries/helpers.py
@@ -1,6 +1,7 @@
 import base64
 import json
 import logging
+import sentry_sdk
 
 from dateutil.parser import parse
 from django.conf import settings
@@ -11,6 +12,7 @@ from json.decoder import JSONDecodeError
 from mail.enums import SourceEnum, ExtractTypeEnum, UnitMapping, ReceptionStatusEnum
 from mail.libraries.email_message_dto import EmailMessageDto, HmrcEmailMessageDto
 from mail.models import LicenceData, UsageData, Mail, GoodIdMapping, LicenceIdMapping
+from mail import serializers
 
 ALLOWED_FILE_MIMETYPES = ["application/octet-stream", "text/plain"]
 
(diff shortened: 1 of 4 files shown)


mail/libraries/routing_controller.py
```diff
@@ -101,6 +106,9 @@ def check_and_route_emails():
             _collect_and_send(pending_message)
 
         logger.info(f"No new emails found from {hmrc_to_dit_server.user} or {spire_to_dit_server.user}")
+
+        publish_queue_status()
+
```
alixedi, line 111: Prolly don't need this given we are calling this at the end of this function?

mail/libraries/helpers.py
```diff
@@ -335,3 +337,72 @@ def get_country_id(country):
...
+        reply_pending_count = reply_pending_qs.count()
+        reply_pending = reply_pending_qs.last()
+        licence_data = LicenceData.objects.get(mail=reply_pending)
+        data["pending_count"] = f"Number of reply pending mails {reply_pending_count} (Expected count is 1)"
+        data[
+            "run_number"
+        ] = f"Current active run number {licence_data.hmrc_run_number} (which is {licence_data.source_run_number} for {licence_data.source})"
+        data["status"] = f"Reply pending, waiting for reply for HMRC run number {licence_data.hmrc_run_number}"
+        if reply_pending.sent_at:
+            print(f"{datetime.now(timezone.utc)} === {reply_pending.sent_at}")
```
alixedi, line 368: Probably don't need a print here.

mail/libraries/helpers.py
```diff
@@ -335,3 +337,72 @@ def get_country_id(country):
...
+    ).order_by("created_at")
+    if reply_pending_qs:
+        reply_pending_count = reply_pending_qs.count()
+        reply_pending = reply_pending_qs.last()
+        licence_data = LicenceData.objects.get(mail=reply_pending)
+        data["pending_count"] = f"Number of reply pending mails {reply_pending_count} (Expected count is 1)"
+        data[
+            "run_number"
+        ] = f"Current active run number {licence_data.hmrc_run_number} (which is {licence_data.source_run_number} for {licence_data.source})"
+        data["status"] = f"Reply pending, waiting for reply for HMRC run number {licence_data.hmrc_run_number}"
```
alixedi, line 366: Arun, the amount of data-wrangling that is happening here makes me wonder if we should just write a serializer?

It will be a nice place to park this kind of field-wise things and we can even write tests for it.

mail/libraries/helpers.py
saruniitr: this is only for debug. Tests failing on CI but not locally so added for that.

mail/libraries/routing_controller.py
saruniitr: No we return from here if there are no new unread mails hence required here also.

mail/libraries/helpers.py
saruniitr: hmm ok, at this point not completely sure what other data we want to export.
I was thinking of going with this and see how it helps with support and extend accordingly.
I will try to add a serializer but most of them will be `SerializerMethodField`

In general, the details exported looks good?

mail/libraries/helpers.py
```diff
@@ -335,3 +337,66 @@ def get_country_id(country):
...
+        if licence_data_qs:
+            licence_data = licence_data_qs.first()
+            queue_status["licence_data"] = serializers.LicenceDataStatusSerializer(licence_data).data
+
+    ud_mail = Mail.objects.filter(extract_type=ExtractTypeEnum.USAGE_DATA).order_by("created_at").last()
+    if ud_mail:
+        usage_data_qs = UsageData.objects.filter(mail=ud_mail)
+        if usage_data_qs:
+            usage_data = usage_data_qs.first()
+            queue_status["usage_data"] = serializers.UsageDataStatusSerializer(usage_data).data
```
saruniitr, line 377: Extra guarding in terms of `if` conditions added because on dev environments we have dummy mailboxes so these `Mail` instances of these types may not present and can cause object not found type of exceptions repeatedly.

mail/libraries/helpers.py
```diff
@@ -335,3 +337,66 @@ def get_country_id(country):
...
+        if ld_mail.status == ReceptionStatusEnum.REPLY_PENDING:
+            queue_status["licence_data"]["waiting_time"] = f"Reply waiting time {processing_time_min} min"
+            if processing_time_min > 30:
+                queue_state = "TO BE MONITORED"
+        elif ld_mail.status == ReceptionStatusEnum.REPLY_SENT:
+            queue_status["licence_data"]["round_trip_time"] = f"Total round trip time {processing_time_min} min"
+
+        # we should only be waiting for reply for one licence data mail
+        if queue_status["licence_data"]["reply_pending_count"] > 1:
+            queue_state = "REQUIRES ATTENTION"
```
saruniitr, line 395: Idea is to give a summary of the state using the above criteria but we can add more conditions going forward and introduce more states if required.

alixedi (commented):
A few more comments - mostly questions.

Also wondering if it would be better to factor out `get_license_data` and `get_usage_data` functions from `publish_queue_status`?

If we can do this, we might also want to include the `processing_time_min` calculation in e.g. `license_data` function to help with the readability.

mail/libraries/helpers.py
```diff
@@ -335,3 +337,66 @@ def get_country_id(country):
...
+    Uses the last Licence data mail and Usage data to get the status of the queue.
+
+    Since we shouldn't sent next licence data mail without receiving reply for the previous one
+    checking the last mail is sufficient to detect any anamolies.
+    """
+    queue_status = {}
+    licence_data = None
+    usage_data = None
+
+    ld_mail = (
```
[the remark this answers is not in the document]
saruniitr: it means `licence_data`

mail/libraries/helpers.py
```diff
@@ -335,3 +337,66 @@ def get_country_id(country):
...
+
+    ud_mail = Mail.objects.filter(extract_type=ExtractTypeEnum.USAGE_DATA).order_by("created_at").last()
+    if ud_mail:
+        usage_data_qs = UsageData.objects.filter(mail=ud_mail)
+        if usage_data_qs:
+            usage_data = usage_data_qs.first()
+            queue_status["usage_data"] = serializers.UsageDataStatusSerializer(usage_data).data
+
+    ## In the increasing order of severity
+    queue_state = "HEALTHY"
```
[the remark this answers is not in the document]
saruniitr: If we don't log to sentry in this case then we won't know the state of the queue without logging into the mailboxes isn't it (when things are running fine)?

mail/libraries/helpers.py
```diff
@@ -335,3 +337,66 @@ def get_country_id(country):
...
+    extra_state = {}
+
+    if licence_data:
+        # usually we should receive reply in 20-30 min
+        processing_time_min = queue_status["licence_data"].pop("processing_time")
+        if ld_mail.status == ReceptionStatusEnum.REPLY_PENDING:
+            queue_status["licence_data"]["waiting_time"] = f"Reply waiting time {processing_time_min} min"
+            if processing_time_min > 30:
+                queue_state = "TO BE MONITORED"
```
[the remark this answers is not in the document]
saruniitr: Taking more than 30min is actually not an issue (atleast from our side) so we just need to monitor updates from hmrc hence went with that state but `RESPONSE TIMEOUT` looks good. I will update this.

mail/libraries/helpers.py
```diff
@@ -335,3 +337,77 @@ def get_country_id(country):
...
+    queue_status["usage_data"] = get_usage_data_status()
+
+    ## In the increasing order of severity
+    queue_state = "HEALTHY"
+    extra_state = {}
+
+    licence_data_status = queue_status.get("licence_data")
+    if licence_data_status:
+        processing_time = licence_data_status.get("processing_time")
+        if "waiting_time" in licence_data_status and processing_time > 30:
```
saruniitr, line 401: this required because of the refactoring as we need to know whether we are waiting for reply or not.

mail/libraries/helpers.py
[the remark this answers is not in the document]
alixedi: Why would we need to know the state of the queue if everything is fine? (not kidding, real question)

alixedi (approved):
GTM!

PR #83, opened 2021-11-18
LTD-1480: Don't send licenceUsage transactions with completion date to LITE-API

saruniitr:
## Change description

Because this is an additional transaction that is sent when all the
allowed quantity of all items on that licence are used. Sending it
again results in double counting of the recorded usage.

The completion date is empty for the actual licenceUsage transaction
which is not disturbed.

### DIFF
--- a/mail/tests/test_send_licence_usage_figures_to_lite_api.py
+++ b/mail/tests/test_send_licence_usage_figures_to_lite_api.py
@@ -1,9 +1,10 @@
 from unittest import mock
+from uuid import uuid4
 
 from rest_framework.status import HTTP_400_BAD_REQUEST, HTTP_207_MULTI_STATUS, HTTP_208_ALREADY_REPORTED
 
 from conf.settings import LITE_API_URL, HAWK_LITE_HMRC_INTEGRATION_CREDENTIALS, LITE_API_REQUEST_TIMEOUT, MAX_ATTEMPTS
-from mail.models import Mail, UsageData, LicencePayload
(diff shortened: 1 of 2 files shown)


mail/tests/test_send_licence_usage_figures_to_lite_api.py
```diff
@@ -175,3 +176,99 @@ def test_schedule_new_task(self, send_licence_usage_figures):
...
+                "4\\usage\\O\\9GB000001328000-PE112345\\R\\20190112\\2\\0\\\\000262\\\\\\\\\n"
+                "5\\end\\line\\5\n"
+                "6\\line\\2\\5\\0\\\n"
+                "7\\usage\\O\\9GB000001328000-PE112345\\R\\20190112\\5\\0\\\\000262\\\\\\\\\n"
+                "8\\end\\line\\5\n"
+                "9\\line\\3\\10\\0\\\n"
+                "10\\usage\\O\\9GB000001328000-PE112345\\R\\20190112\\10\\0\\\\000262\\\\\\\\\n"
+                "11\\end\\line\\5\n"
+                "12\\end\\licenceUsage\\10\n"
+                "13\\licenceUsage\\LU04148/00002\\insert\\GBSIEL/2020/0000025/P\\C\\20211110\n"
```
saruniitr, line 212: `20211110` is the completion date and also `C` means cancelled here
This transaction needs to be filtered out and not send to LITE-API to avoid double counting

alixedi (approved):

PR #86, opened 2022-01-21
Some ideas on generating EDIFACT from license

alixedi:
ATM, the EDIFACT is generated from licenses in batches. If there are any errors in any of the licenses, the whole batch fails. This landed us into some issues lately and we had to work around this with a [bit of a hack](https://github.com/uktrade/lite-hmrc/pull/85).

The hack was put in because the code as it currently stands is not amenable to processing licenses one by one. This PR has some ideas that I was working on (before I gave up and put in the hack).

### DIFF
--- a/mail/libraries/lite_to_edifact_converter.py
+++ b/mail/libraries/lite_to_edifact_converter.py
@@ -15,7 +15,9 @@ from mail.libraries.edifact_validator import (
 
 
 class EdifactValidationError(Exception):
-    pass
+    def __init__(self, message, errors):
+        super().__init__(message)
+        self.errors = errors
 
 
 def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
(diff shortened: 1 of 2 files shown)


mail/libraries/lite_to_edifact_converter.py
```diff
@@ -15,7 +15,9 @@
 
 
 class EdifactValidationError(Exception):
-    pass
+    def __init__(self, message, errors):
+        super().__init__(message)
+        self.errors = errors
```
alixedi, line 20: Attach the errors with the Exception so that when you `raise`, you get the right context in Sentry.

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -155,13 +157,20 @@ def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
...
 
     return edifact_file
 
 
+def process_edifact_line(line, line_no):
+    """Tries to validate a line before adding it to the EDIFACT file.
+    If the validation fails, skips the line so that we can process the
+    rest of the licenses.
+    """
+    pass
```
alixedi, line 172: I started writing this but there is more to it...

1) We should probably have a function called `get_license_edifact`. This should create all the EDIFACT lines for a particular license and called inside the loop. 

2) `get_license_edifact` should validate the EDIFACT blob produced for each license. If it cannot produce a valid blob, it should raise an exception. On the other hand, if it produces a valid blob, the should mark the license as `processed` **inside the loop**. instead of the following (up the stack) -

https://github.com/uktrade/lite-hmrc/blob/dcccd1cec37acbeb9910ca47e4f529baed41f2a4/mail/libraries/builders.py#L134-L143

PR #87, opened 2022-01-25
LTD-1336: Add management command to resend an email for a given run number

saruniitr:
## Change description

One of the support task is to manually resend an email to either HMRC/SPIRE.
This involves logging into the mailbox, identify the correct mail using the
run number, downloading the attachment and resending it. This is not efficient
and error prone so an management command is added to pick up the correct
mail as per the given run number and resend it.

Depending on the run number, mail extract type and status the command determines
the destination and sends it again.


mail/libraries/builders.py
```diff
@@ -63,6 +63,57 @@ def build_request_mail_message_dto(mail: Mail) -> EmailMessageDto:
...
+        spire_data, _ = split_edi_data_by_id(mail.edi_data)
+        if len(spire_data) > 2:  # if SPIRE blocks contain more than just a header & footer
+            file = build_edifact_file_from_data_blocks(spire_data)
+            attachment = [
+                build_sent_filename(mail.edi_filename, run_number),
+                build_sent_file_data(file, run_number),
+            ]
+
+    logging.info(
+        f"Preparing request Mail dto of extract type {mail.extract_type}, sender {sender}, receiver {receiver} with filename {attachment[0]}"
```
kevincarrogan, line 102: You shouldn't use f-strings with logging output.

It's also better to use `logger = logging.getLogger(__name__)`.

mail/libraries/builders.py
```diff
@@ -63,6 +63,57 @@ def build_request_mail_message_dto(mail: Mail) -> EmailMessageDto:
...
+        sender = settings.EMAIL_USER
+        receiver = settings.SPIRE_ADDRESS
+        update = UsageData.objects.get(mail=mail)
+        run_number = update.spire_run_number
+        spire_data, _ = split_edi_data_by_id(mail.edi_data)
+        if len(spire_data) > 2:  # if SPIRE blocks contain more than just a header & footer
+            file = build_edifact_file_from_data_blocks(spire_data)
+            attachment = [
+                build_sent_filename(mail.edi_filename, run_number),
+                build_sent_file_data(file, run_number),
```
kevincarrogan, line 98: Are these three the only extract_type cases? What happens if another extract type gets pushed through? Should this fail?

mail/management/commands/resend_email.py
```diff
@@ -0,0 +1,126 @@
...
+    Case3: When a usage_data mail is sent to SPIRE successfully from our system but it hasn't reached SPIRE
+    In this case we expect status of the mail to be 'reply_sent' and extract_type as 'usage_data'
+
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--hmrc_run_number", type=str, nargs="?", help="Run number used by HMRC to identify the mail to be resent"
+        )
+        parser.add_argument("--dry_run", type=str, nargs="?", help="Is it a test run?", default="True")
```
kevincarrogan, line 76: This shouldn't be a string.

Do:
```
parser.add_argument("--dry-run", help="Is it a test run?", action="store_true")
```

mail/libraries/builders.py
saruniitr: there is one more type `usage_reply` but those mails are not sent to HMRC so not required, anything else we bail out early.

mail/management/commands/resend_email.py
saruniitr: :+1:

mail/libraries/builders.py
kevincarrogan: I think I'm missing where we bail out if it's not one of these three types. To me it looks like we still. attempt to return the EmailMessageDto or is something happening in the caller of this function to stop it doing anything with another extract_type?

saruniitr: we bail out early in the `handle()` itself before calling this function but I can additional checks in this function also.

kevincarrogan (approved):

PR #88, opened 2022-01-31
Ltd 1746 failed to send usage

kevincarrogan:


mail/tests/test_usage_data_decompostion.py
```diff
@@ -236,6 +236,77 @@ def test_usage_data_split_according_to_licence_ids(self):
         self.assertEqual(spire_data, spire_data_expected)
         self.assertEqual(lite_data, lite_data_expected)
 
+    @tag("1022", "carriage-return-line-endings")
```
saruniitr, line 239: Looks like previously tag is used to indicate the ticket number which is not correct for this one, you might as well remove it completely.

saruniitr (approved):

PR #94, opened 2022-02-03
Update logging calls to add stack trace appropriately

kevincarrogan:


alixedi (approved):

mail/libraries/usage_data_decomposition.py
```diff
@@ -113,12 +113,10 @@ def build_json_payload_from_data_blocks(data_blocks: list) -> dict:
...
-        """
-        Filter blocks that has a completion date because this is an additional transaction
-        that is sent when all the allowed quantity of all items on that licence are used.
-        Sending it again results in double counting of the recorded usage
-        """
-        payload = [item for item in payload if not item["completion_date"]]
+    # Filter blocks that has a completion date because this is an additional transaction
+    # that is sent when all the allowed quantity of all items on that licence are used.
+    # Sending it again results in double counting of the recorded usage
+    payload = [item for item in payload if not item["completion_date"]]
```
saruniitr, line 119: I think this should be within the for loop

kevincarrogan: The payload is a list of items that's being appended to on each run of the loop. At the end we then remove all of the completion_date items, if you do it in the loop you're just doing the same thing over and over again.

saruniitr (approved):

PR #99, opened 2022-03-02
ICMSLST-1464 Fix broken use of logging

davidwtbuxton:
These logging calls had a mix of f-strings and format calls, which "works"
but does not do what was intended.

    >>> f"Status {0}".format("fail")
    'Status 0'

But for Python's logging module, one should use positional arguments, not
string formatting (f-strings or other flavours). This way the logging
module can delay string formatting until it is needed (or even never if
the logging level means the message is skipped). Also the logging module
will swallow unicode encode/decode errors when formatting strings, which
allows the program to continue.

### DIFF
--- a/mail/management/commands/resend_email.py
+++ b/mail/management/commands/resend_email.py
@@ -26,13 +26,13 @@ def get_mail_extract(hmrc_run_number):
         licence_data = LicenceData.objects.get(hmrc_run_number=hmrc_run_number)
         mail = licence_data.mail
     except LicenceData.DoesNotExist:
-        logger.info("No licence data instance found for given run number {0}".format(hmrc_run_number))
+        logger.info("No licence data instance found for given run number %s", hmrc_run_number)
 
     try:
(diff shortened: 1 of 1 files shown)


saruniitr (approved):

davidwtbuxton:
I got well confused because the CI tests wouldn't run when I first opened this PR (because I didn't have permissions for this repo) but then did run when I pushed the branch with this commit to the repo directly.

But all good now.

davidwtbuxton:
There are a lot of places in the code which using logging with f-strings. I haven't updated those, but it would be nice to make those use logging correctly, i.e. using positional arguments and not using formatting strings.

https://docs.python.org/3/howto/logging.html#logging-variable-data

saruniitr:
> There are a lot of places in the code which using logging with f-strings. I haven't updated those, but it would be nice to make those use logging correctly, i.e. using positional arguments and not using formatting strings.
> 
> https://docs.python.org/3/howto/logging.html#logging-variable-data

:+1:

kevincarrogan (approved):

PR #105, opened 2022-03-14
Skip reporting usage data when there are no licences

saruniitr:
## Change description

HMRC sends us usage data when a licence is used but when the licence is
completed i.e., all the allowed quantity of all products on that licence
are used then we get another usage data transaction with licence status
as cancelled. Currently we skip these transactions hence the licence count
in the payload comes up as 0 but we are trying to report this to LITE.

Since the licences are empty, LITE considers this as an error and throws
an exception. This is not an error and we can skip reporting in this case.

### DIFF
--- a/mail/tasks.py
+++ b/mail/tasks.py
@@ -57,10 +57,23 @@ def send_licence_usage_figures_to_lite_api(lite_usage_data_id):
         )
         return
 
+    # Extract usage details of Licences issued from LITE
+    _, data = split_edi_data_by_id(lite_usage_data.mail.edi_data, lite_usage_data)
+    payload = build_json_payload_from_data_blocks(data)
+
+    # We only process usage data for active licences so below error is unlikely
+    if len(payload["licences"]) == 0:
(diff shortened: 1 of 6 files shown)


mail/tasks.py
```diff
@@ -57,10 +57,30 @@ def send_licence_usage_figures_to_lite_api(lite_usage_data_id):
...
+    Each UsageData transaction contains an action and licence status.
+    To extract usage details we only need to process the transactions with
+    action as 'insert' and licence status as 'O (open)', because in some cases
+    we get multiple transactions for the same licence reference.
+    Eg when all the approved quantity of all goods on the licence are shipped
+    then we get a transaction with status 'C (cancelled)' which gets filtered
+    while building the payload and licence_count in this case will be 0, so
+    we don't need to report usage in this case
+    """
+    if len(payload["licences"]) == 0:
```
saruniitr, line 73: currently we are trying to report in this case also and this is one of the reason for many such alerts in the channel

kevincarrogan (approved):

PR #109, opened 2022-03-17
LTD: Remove unnecessary mail is pending error as it is not an error

saruniitr:
## Change description

When send_licence_data_to_hmrc() task runs it checks to see if there is an update
in progress. An update can be either we are waiting for a reply or the next email
is about to be processed (eg new email just got created and check_inbox task about
to process it, in this case it will be in pending state).

Since we have to wait till we completely process an email (i.e, reply_sent state)
we should skip this update when there is an update in progress, this gets picked
up in the next polling cycle.

Currently we are logging an error if an update in progress. This is not an error
and it unnecessarily adds noise to the alerts hence remove this.

### DIFF
--- a/mail/tasks.py
+++ b/mail/tasks.py
@@ -14,7 +14,7 @@ from django.db import transaction
 from django.utils import timezone
 from rest_framework.status import HTTP_207_MULTI_STATUS, HTTP_208_ALREADY_REPORTED
 
-from mail.enums import ReceptionStatusEnum, ReplyStatusEnum
+from mail.enums import ReceptionStatusEnum
 from mail.libraries.builders import build_licence_data_mail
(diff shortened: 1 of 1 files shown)


mail/tasks.py
```diff
@@ -244,31 +246,7 @@ def send_licence_data_to_hmrc():
...
-        return False
-
-    return True
-
-
-def _get_pending_mail() -> []:
-    return list(Mail.objects.exclude(status=ReceptionStatusEnum.REPLY_SENT).values_list("id", flat=True))
-
-
-def _get_rejected_mail() -> []:
```
saruniitr, line 260: unused hence deleted

wkeeling (approved):

PR #114, opened 2022-03-30
ICMSLST-1486 Count licences using the CHIEF message

davidwtbuxton:
Instead of making SQL queries, use the actual count of "licence"
transactions that are present in a CHIEF message (which is what
that trailing count is supposed to be).

The advantage of this is we can untangle some of the database
logic from building the actual CHIEF message.

### DIFF
--- a/mail/libraries/lite_to_edifact_converter.py
+++ b/mail/libraries/lite_to_edifact_converter.py
@@ -199,7 +199,7 @@ def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
 
     # File trailer includes the number of licences, but +1 for each "update"
     # because this code represents those as "cancel" followed by "insert".
-    num_transactions = licences.count() + licences.filter(action=LicenceActionEnum.UPDATE).count()
+    num_transactions = chiefprotocol.count_transactions(lines)
     file_trailer = ("fileTrailer", num_transactions)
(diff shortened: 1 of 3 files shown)


mail/libraries/lite_to_edifact_converter.py
```diff
@@ -199,7 +199,7 @@ def licences_to_edifact(licences: QuerySet, run_number: int) -> str:
 
     # File trailer includes the number of licences, but +1 for each "update"
     # because this code represents those as "cancel" followed by "insert".
-    num_transactions = licences.count() + licences.filter(action=LicenceActionEnum.UPDATE).count()
+    num_transactions = chiefprotocol.count_transactions(lines)
```
saruniitr, line 202: the comment above talks about adding `+1` for `update`, does `count_transactions()` cover that as well?

davidwtbuxton: Yes. There is an existing test for an update request which also shows that count = 2 for 1 update.  

https://github.com/uktrade/lite-hmrc/blob/master/mail/tests/test_licence_to_edifact.py#L100

Interestingly the spec says the transaction count is optional. Would be nice if we omitted it altogether and everything still works.

saruniitr (approved):

PR #116, opened 2022-03-31
ICMSLST-1523 Yield an end line with others when generating CHIEF mess…

davidwtbuxton:
…ages

And added tests to cover how country lines are generated.

### DIFF
--- a/mail/tests/test_licence_to_edifact.py
+++ b/mail/tests/test_licence_to_edifact.py
@@ -5,6 +5,7 @@ from parameterized import parameterized
 
 from mail.enums import LicenceActionEnum
 from mail.libraries.lite_to_edifact_converter import (
+    generate_lines_for_licence,
     licences_to_edifact,
     get_transaction_reference,
     EdifactValidationError,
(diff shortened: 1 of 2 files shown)


mail/tests/test_licence_to_edifact.py
```diff
@@ -172,7 +173,46 @@ def test_foreign_trader_address_sanitize(self, address_line_1, expected_trader_l
...
+        }
+        licence = LicencePayload(reference="GBSIEL/123", data=data)
+        lines = list(generate_lines_for_licence(licence))
+
+        # Note there are 2 country lines.
+        expected_types = ["licence", "trader", "country", "country", "restrictions", "line", "end"]
+        self.assertEqual([line[0] for line in lines], expected_types)
+        # The country code is the 2nd field.
+        self.assertEqual(lines[2], ("country", "GB", None, "D"))
+        self.assertEqual(lines[3], ("country", "NIR", None, "D"))
```
saruniitr, line 218: I understand this is an example but country code is only two characters as per the spec.

davidwtbuxton: OK cool, I took `NIR` from a list of country codes somewhere in the ICMS code base. I will change this to `NI`.

saruniitr (approved):

PR #119, opened 2022-04-07
ICMSLST-1541 Add types for CHIEF message lines

davidwtbuxton:
This uses dataclasses to define the line structure, because they support
inheriting field declarations whereas collections.namedtuple does not.
Dataclasses also allow using keyword arguments with the constructor.

The big improvement is we can have a little bit more self-documenting
code with names for the fields, instead of items in tuples with an
order that seems arbitrary but is very significant.

In CHIEF syntax every line starts with the line number and record type,
but for our use it is very useful to leave the line number unknown
until we construct the final message. But if you have one field in a
dataclass as optional, then all fields that follow must also be optional.

### DIFF
--- a/mail/libraries/chieftypes.py
+++ b/mail/libraries/chieftypes.py
@@ -1,0 +1,165 @@
+from dataclasses import dataclass
+from typing import Optional
+
+# These types represent the structure of records in a CHIEF message. A message
+# is one of licenceData, licenceReply, usageData, or usageReply types.
+
+# Every line in a message starts with a line number, and then the record type.
+@dataclass
+class _Record:
+    lineno: Optional[int] = None
+    type_: Optional[str] = None
+
+
+# Every message starts and ends with the file header.
+@dataclass
+class FileHeader(_Record):
+    type_: Optional[str] = "fileHeader"
+    source_system: Optional[str] = None
+    destination_system: Optional[str] = None
+    data_id: Optional[str] = None
+    creation_date_time: Optional[str] = None
+    run_num: Optional[int] = None
+    reset_run_num: Optional[str] = None
+
+
+@dataclass
+class FileTrailer(_Record):
+    type_: Optional[str] = "fileTrailer"
+    transaction_count: Optional[int] = None
+    # The spec also allows a hash_total field, but this implementation
+    # flags that as an error when validating.
+    # hash_total: Optional[str] = None
+
+
+# Several types have a matching "end" line.
+@dataclass
+class End(_Record):
+    type_: Optional[str] = "end"
+    start_record_type: Optional[str] = None
+    record_count: Optional[int] = None
+
+
+# Licence data (request).
+@dataclass
+class Licence(_Record):
+    type_: Optional[str] = "licence"
+    transaction_ref: Optional[str] = None
+    action: Optional[str] = None
+    licence_ref: Optional[str] = None
+    licence_type: Optional[str] = None
+    usage: Optional[str] = None
+    start_date: Optional[str] = None
+    end_date: Optional[str] = None
+
+
+@dataclass
+class Trader(_Record):
+    type_: Optional[str] = "trader"
+    turn: Optional[str] = None
+    rpa_trader_id: Optional[str] = None
+    start_date: Optional[str] = None
+    end_date: Optional[str] = None
+    name: Optional[str] = None
+    address1: Optional[str] = None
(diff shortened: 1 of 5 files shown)


mail/libraries/chieftypes.py
```diff
@@ -0,0 +1,165 @@
...
+@dataclass
+class _Record:
+    lineno: Optional[int] = None
+    type_: Optional[str] = None
+
+
+# Every message starts and ends with the file header.
+@dataclass
+class FileHeader(_Record):
+    type_: Optional[str] = "fileHeader"
```
saruniitr, line 17: Could you clarify the significance of `_` at the end of `type`?
Similarly in `_Record` it signifies for internal use?

mail/libraries/chieftypes.py
```diff
@@ -0,0 +1,165 @@
...
+    source_system: Optional[str] = None
+    destination_system: Optional[str] = None
```
saruniitr, line 19: These are not optional in this record and many other fields in the other classes as well especially we have validation for these records so may be enough to define them as `str` but not marking them as `Optional`?

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -35,92 +35,86 @@ def generate_lines_for_licence(licence):
...
+            address4=trader.get("address").get("line_4", ""),
+            address5=trader.get("address").get("line_5", ""),
+            postcode=trader.get("address").get("postcode"),
         )
 
         # Uses "D" for licence use because lite only sends allowed countries, to use E would require changes on the API
         if payload.get("type") in LicenceTypeEnum.OPEN_LICENCES:
             if payload.get("country_group"):
-                yield ("country", None, payload.get("country_group"), "D")
+                yield chieftypes.Country(group=payload.get("country_group"), use="D")
```
saruniitr, line 88: not related to this PR but we should probably define constants for `"D"` at the end

mail/libraries/chieftypes.py
davidwtbuxton: This field is named `type_` so that it does not shadow the built-in `type`.

The underscore at the start of `_Record` is to indicate the class should be considered an internal implementation detail, not used more generally. I wanted a base class for these CHIEF message line types so that we can build a registry of all the line types. This will be useful for validating messages (not implemented in this patch).

mail/libraries/chieftypes.py
davidwtbuxton: I marked all these as optional because the record types are used to represent a CHIEF message line, but I want to allow the first field (the line number) to be optional so that we can assign the correct line number when going to build the final message.

It is also useful to mark other fields optional since as far as I know Python's type system cannot represent interconnected field dependencies (e.g. make C required when A = foo, but B required when A = bar) but even if it could, this project doesn't begin to enforce Python typing.

Marking a type as optional (allowing `None`) is useful even when the message protocol requires a valid identifier.

mail/libraries/lite_to_edifact_converter.py
davidwtbuxton: Yes, either via `typing.Literal` or an enum. An enum probably makes more sense since there is no type checking in this project.

https://docs.python.org/3/library/typing.html#typing.Literal
https://docs.python.org/3/library/enum.html

mail/libraries/chieftypes.py
saruniitr: Actually I don't mean from type dependencies point of view, I mean eg for this record `source` and `destination` etc are required, without them edifact validation fails when generating the file, similarly for other fields in other classes. So just wondering if marking them as optional causes any confusion.

davidwtbuxton: I want to add comments with a lot of the detail that is in the CHIEF message spec about the allowed fields and the meaning of those fields. That would help clarify that `Optional` here refers to Python typing.

[This page links to various parts of the specs][1], but not the part that defines the CHIEF message.

[1]: https://www.gov.uk/government/collections/chief-user-guides-and-technical-specifications

mail/libraries/chieftypes.py
saruniitr: It just makes it slightly difficult as it is easy to miss typing the `_` at the end for type, hopefully it is not going to be an issue.
With `_Record` it makes sense as it is for internal use and not expected to be used too often.

saruniitr (approved):

PR #122, opened 2022-04-08
ICMSLST-1544 Apply isort code style to Python `import ..` statements

davidwtbuxton:
And after this is merged we can add isort to the pre-commit hook and to the CircleCI config.


kevincarrogan:
Maybe this would be good to add as a `pre-commit` hook as well? In fact then the linting could all just be done by `pre-commit` e.g. by running `pre-commit run -a` (I think that's the command at least) or just letting `pre-commit` pick it up when committing.

kevincarrogan:
> Maybe this would be good to add as a `pre-commit` hook as well? In fact then the linting could all just be done by `pre-commit` e.g. by running `pre-commit run -a` (I think that's the command at least) or just letting `pre-commit` pick it up when committing.

Just realised it _is_ added as a pre-commit hook. Then we can just change the README to say to apply linting then just use `pre-commit run -a`.

kevincarrogan (approved):

PR #124, opened 2022-04-11
ICMSLST-1545 Remove unused code

davidwtbuxton:
- Tests that were never run, and which fail anyway.
- Unused API views.
- Unused helper methods on the enum-like classes.
- Unused base64 helpers.

### DIFF
--- a/mail/views.py
+++ b/mail/views.py
@@ -8,9 +8,9 @@ from rest_framework.views import APIView
 
 from conf.authentication import HawkOnlyAuthentication
 from mail.enums import LicenceActionEnum, LicenceTypeEnum, ReceptionStatusEnum
-from mail.models import LicenceData, LicenceIdMapping, LicencePayload, Mail, UsageData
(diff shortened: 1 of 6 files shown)


mail/views.py
```diff
@@ -82,12 +82,6 @@ def post(self, request):
             )
 
 
-class ManageInbox(APIView):
```
saruniitr, line 85: This one and the view below are used by integration tests in `lite-api` so we don't want to remove them.

davidwtbuxton: I don't see where the  `ManageInbox` and `SendUsageUpdatesToLiteApi` views are exposed in the URL configuration, which is why I removed them. How do the lite-api tests invoke them?

saruniitr: if they are not exposed probably safe to remove them. I need to check how the tests invoke them.

saruniitr (approved):

PR #125, opened 2022-04-14
ICMSLST-1549 More info on the meaning of CHIEF message fields

davidwtbuxton:
These comments are adapted from the DES 235 and DES 236 specifications.

### DIFF
--- a/mail/libraries/chieftypes.py
+++ b/mail/libraries/chieftypes.py
@@ -15,11 +15,26 @@ class _Record:
 @dataclass
 class FileHeader(_Record):
     type_: Optional[str] = "fileHeader"
+
+    # System that generates the input file. Required.
     source_system: Optional[str] = None
+
+    # System that processes the input transactions. Required.
     destination_system: Optional[str] = None
     data_id: Optional[str] = None
+
+    # Date-time of creation of the input file, format ccyymmddhhmm.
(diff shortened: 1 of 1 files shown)


mail/libraries/chieftypes.py
```diff
@@ -15,11 +15,25 @@ class _Record:
...
+
+    # The run number must be 1 on the first transmission and then incremented
+    # (by one) on each subsequent transmission. A run number is not expected to
+    # cycle but if it can for an interface it should roll to zero.
     run_num: Optional[int] = None
+
+    # If "Y" the given runNum overrides the value expected by the processing
+    # system. For sequential processing, the next file sent should have the
+    # next number. For parallel processing it must be set to "Y".
     reset_run_num: Optional[str] = None
```
saruniitr, line 37: Almost always we want this to be `N`. Could you add this note here? I think we hardcode to `N` during generation anyway iirc.

saruniitr (approved):

mail/libraries/chieftypes.py
davidwtbuxton: Updated with `For LITE / SPIRE this is usually (always?) "N".`. How's that?

saruniitr: sounds good to me

PR #126, opened 2022-04-19
ICMSLST-1537 Use gunicorn for serving in production

davidwtbuxton:
Instead of Django's development server, use gunicorn. Default behaviour
reads a configuration from gunicorn.conf.py and binds to 0.0.0.0:$PORT
when $PORT is defined in the environment.

The config overrides the `Server` response header to hide gunicorn's
signature. This is silly, but part of a belt-and-braces approach to
security. However the header is then reset by another proxy. If we ever
deploy the app with a proxy that respects the app `Server` header, it
would show the customised, obscured signature.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
-web: python manage.py migrate && ./manage.py runserver 0.0.0.0:$PORT
+# See gunicorn.conf.py for more configuration.
+web: python manage.py migrate && gunicorn conf.wsgi:application
 worker: python manage.py process_tasks --log-std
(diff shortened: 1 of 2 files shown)


davidwtbuxton:
N.B. the gunicorn setup here is super basic, using the default sync (fork) worker. This is different to the recommendation to use the gevent worker and patch various things to be async friendly.

kevincarrogan (approved):

saruniitr (approved):

PR #131, opened 2022-05-03
ICMSLST-1547 Backport Django's Choices/enum type for use with enums

davidwtbuxton:
Django version 3.x added the choice types for use as enums, so I backported
that implementation and used it for the enum things.

PR #132, opened 2022-05-10
    ICMSLST-1570 Use positional arguments to logging calls

davidwtbuxton:
The stdlib's logging library is intended to be used with positional arguments,
so that a call that doesn't emit a message can skip formatting.

There were also some totally useless logging calls that have been removed.


davidwtbuxton:
This PR includes changes from #130 and #131 so those should be merged before this.

kevincarrogan (approved):

saruniitr (approved):

PR #137, opened 2022-05-25
ICMSLST-1627 Catch invalid licence data

davidwtbuxton:
The CHIEF message serializer requires certain fields have certain values,
but the API was not enforcing that.

- goods/unit must be one of the known values.
- goods/id must be a UUID value.

### DIFF
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -179,11 +179,9 @@ class UnitMapping(enum.Enum):
     ITG = 30  # intangible
 
     @classmethod
-    def convert(cls, unit):
-        try:
-            return cls[unit].value
-        except KeyError:
-            pass
+    def serializer_choices(cls):
+        # Used by the API serializer for validation.
+        return list(cls.__members__.keys())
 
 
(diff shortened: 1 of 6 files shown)


davidwtbuxton:
I added `GoodSerializer.id` as an _optional_ field. [The message construction stuff in generate_lines_for_licence][1] appears to assume that field is present in the submitted licence data, but I wasn't sure if that is always the case:

```
        if payload.get("goods") and payload.get("type") in LicenceTypeEnum.STANDARD_LICENCES:
            for g, commodity in enumerate(payload.get("goods"), start=1):
                GoodIdMapping.objects.get_or_create(
                    lite_id=commodity["id"], licence_reference=licence.reference, line_number=g
                )
```

[1]: https://github.com/uktrade/lite-hmrc/blob/1076b8e3c37bdf4176bb1634c74cfa8d26f9c0bc/mail/libraries/lite_to_edifact_converter.py#L119

saruniitr (approved):

PR #148, opened 2022-06-20
ICMSLST-1654 Add ICMS serializer for FA-OIL application.

MattHolmes123:
Changes:

- Add ICMS serializer for FA-OIL application.
- Add models to admin interface for debugging.
- Fix optional date values throwing exceptions.

### DIFF
--- a/mail/icms_serializers.py
+++ b/mail/icms_serializers.py
@@ -1,0 +1,66 @@
+import re
+
+from rest_framework import serializers
+
+ICMS_CASE_REF_PATTERN = re.compile(
+    r"""
+        ^       # Start of string
+        [a-z]+  # Prefix
+        /       # Separator
+        \d+     # Year
+        /       # Separator
+        \d+     # reference
+        (/\d+)? # Optional variation request
+        $       # End of string
+    """,
+    flags=re.IGNORECASE | re.VERBOSE,
+)
+
+
+class AddressSerializer(serializers.Serializer):
+    line_1 = serializers.CharField(max_length=35)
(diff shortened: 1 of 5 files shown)


mikeylockwood (approved):

mail/icms_serializers.py
```diff
@@ -0,0 +1,66 @@
...
+)
+
+
+class AddressSerializer(serializers.Serializer):
+    line_1 = serializers.CharField()
+    line_2 = serializers.CharField(required=False)
+    line_3 = serializers.CharField(required=False)
+    line_4 = serializers.CharField(required=False)
+    line_5 = serializers.CharField(required=False)
+    postcode = serializers.CharField()
```
saruniitr, line 26: I see there is a `max_length` defined for `reference` in a serializer below, there are length restrictions for address also, do we need similar here?

MattHolmes123: We may do but they don't exist in the lite version either.
I will update the ICMS versions based on this info:
image

saruniitr: correct, they are not in LITE but that is checked when generating email message
https://github.com/uktrade/lite-hmrc/blob/master/mail/libraries/edifact_validator.py#L182

saruniitr (approved):

mail/icms_serializers.py
MattHolmes123: yer I noticed all that validation.
I guess I'd rather validate the incoming data rather than going to the effort of saving / processing it first?
Can't harm having it in both places

PR #149, opened 2022-06-24
Fix hawk authentication

MattHolmes123:
Changes:

- Return correct exception when authentication fails for any reason
- Add authentication classes to app endpoints (even if they are only used in testing)
- Add local_settings sample to reduce logging output (remove json formatter).

### DIFF
--- a/conf/authentication.py
+++ b/conf/authentication.py
@@ -46,7 +46,7 @@ def _authenticate(request):
     Raises a HawkFail exception if the passed request cannot be authenticated
     """
 
-    if settings.HAWK_AUTHENTICATION_ENABLED:
+    if hawk_authentication_enabled():
         return Receiver(
             _lookup_credentials,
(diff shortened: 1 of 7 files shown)


conf/authentication.py
```diff
@@ -4,26 +4,42 @@
...
-            raise e
+            logger.warning("Failed HAWK authentication %s", e)
+
+            raise exceptions.AuthenticationFailed(f"Failed HAWK authentication")
+
+        except Exception as e:
+            logger.error("Failed HAWK authentication %s", e)
+
+            if settings.SENTRY_ENABLED:
+                capture_exception(e)
```
saruniitr, line 34: My understanding is exceptions are automatically captured by sentry, do we need to this explicitly? how is this different?

MattHolmes123: It sends the unknown exception to sentry but then gracefully returns a 401 response to the caller.

PR #154, opened 2022-07-12
LTD-2508: Upgrade Django to current LTS version 3.2

saruniitr:
## Change description

The support for the current Django version has ended in April 2022
hence upgrading to current LTS version 3.2.

Chosen version is `3.2.14` which is the same as `lite-api` version.

**Other changes**:
- Removed usages of `django-jsonfield` as Django has inbuilt support for json from 3.1. This package still exists in `Pipfile` but not used because initial migrations need this.
- `elastic-apm` is also upgraded to the same version as in `lite-api` as the deployments were failing with the upgrade
- Add tests for `Healthcheck` view to satisfy `codecov`


MattHolmes123:
@saruniitr Do you plan on removing this dependency now you have upgraded django:
https://pypi.org/project/django-jsonfield/
image

Also models.TextChoices was available from Django3.0 so we should look to getting rid of that backport.
All of this:
https://github.com/uktrade/lite-hmrc/blob/master/mail/enums.py#L8-L89

mail/migrations/0018_auto_20220712_1452.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('mail', '0017_auto_20220517_0830'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
currycoder, line 13: What SQL will actually run for this alter field operation?  Or is it likely to have no SQL operation?  I'm assuming that the old jsonfield library stored this as a postgres json field...

conf/views.py
```diff
@@ -83,9 +83,9 @@ def _get_rejected_mail() -> []:
...
     @staticmethod
-    def _build_response(status, message, start_time):
+    def _build_response(request, status, message, start_time):
         duration_ms = (time.time() - start_time) * 1000
         response_time = "{:.3f}".format(duration_ms)
-        context = {"message": message, "response_time": response_time}
+        context = {"message": message, "response_time": response_time, "status": status}
 
-        return render_to_response("healthcheck.xml", context, content_type="application/xml", status=status)
+        return render(request, "healthcheck.xml", context, content_type="application/xml", status=status)
```
currycoder, line 91: I think you could use `self.request` rather than having to pass request round in the function parameters everywhere.  Could be worth a try?

mail/models.py
```diff
@@ -109,21 +108,21 @@ def get_licence_ids(self):
 
 class UsageData(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    licence_ids = JSONField()
+    licence_ids = models.JSONField(default=dict)
```
currycoder, line 111: *should* all of these be `default=dict`?  licence_ids for example I could see as being a list, but my knowledge of the data is limited

mail/migrations/0018_auto_20220712_1452.py
saruniitr: I think no SQL operation.
Correct previous library stored it as json field but after the upgrade it was coming out as `str`

conf/views.py
saruniitr: :+1:

mail/models.py
saruniitr: @currycoder I have corrected few other fields also.

mail/migrations/0018_auto_20220712_1452.py
MattHolmes123: You can get django to show you the sql btw:
https://docs.djangoproject.com/en/4.0/ref/django-admin/#sqlmigrate

PR #155, opened 2022-08-04
Separate out authentication from mail server

kevincarrogan:
This change is in preparation for when we add OAuth authentication, this separates out the concerns for authenticating the connection and allows us to swap out authentication on a case by case basis if necessary

### DIFF
--- a/mail/auth.py
+++ b/mail/auth.py
@@ -1,0 +1,23 @@
+import poplib
+
+from typing_extensions import Protocol
+
+
+class Authenticator(Protocol):
+    user: str
+
+    def authenticate(self, connection: poplib.POP3_SSL):
+        ...
+
+
+class BasicAuthentication:
+    def __init__(self, user: str, password: str):
+        self.user = user
+        self.password = password
+
+    def authenticate(self, connection: poplib.POP3_SSL):
+        connection.user(self.user)
+        connection.pass_(self.password)
+
+    def __eq__(self, other: Authenticator):
+        return self.user == other.user and self.password == other.password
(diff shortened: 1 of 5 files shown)


mail/auth.py
```diff
@@ -0,0 +1,23 @@
+import poplib
+
+from typing_extensions import Protocol
+
+
+class Authenticator(Protocol):
+    user: str
+
+    def authenticate(self, connection: poplib.POP3_SSL):
+        ...
```
currycoder, line 10: I've not seen this before... Shouldn't it be `pass`?

mail/auth.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class BasicAuthentication:
+    def __init__(self, user: str, password: str):
+        self.user = user
+        self.password = password
+
+    def authenticate(self, connection: poplib.POP3_SSL):
+        connection.user(self.user)
+        connection.pass_(self.password)
```
currycoder, line 20: What an odd interface

mail/libraries/routing_controller.py
```diff
@@ -48,28 +53,40 @@ def get_hmrc_to_dit_mailserver() -> MailServer:
...
+        auth,
+        hostname=settings.HMRC_TO_DIT_EMAIL_HOSTNAME,
         pop3_port=settings.HMRC_TO_DIT_EMAIL_POP3_PORT,
     )
 
 
 def get_mock_hmrc_mailserver() -> MailServer:
-    return MailServer(
-        hostname=settings.MOCK_HMRC_EMAIL_HOSTNAME,
+    auth = BasicAuthentication(
```
currycoder, line 69: Is it worth adding some simple tests to cover these lines?

currycoder (approved):
Looks good, just a few small questions which I will leave with you

mail/auth.py
kevincarrogan: https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols

PR #158, opened 2022-08-18
Add modern authentication auth class

kevincarrogan:
This adds an authentication class for OAuth based authentication for pop3 connections.

At this point in time this is unused but is laying the ground work for the switch to OAuth.

### DIFF
--- a/mail/auth.py
+++ b/mail/auth.py
@@ -54,8 +54,8 @@ class ModernAuthentication:
     def _get_access_token(self):
         scopes = ["https://outlook.office.com/.default"]
 
-        result = self.app.acquire_token_silent(scopes, account=None)
-        if not result:
+        result = self.app.acquire_token_silent(scopes, account=None)  # This attempts to get the token from the cache
+        if not result:  # If we don't find the token in the cache then we go off and retrieve it from the provider
             result = self.app.acquire_token_for_client(scopes=scopes)
 
         return result["access_token"]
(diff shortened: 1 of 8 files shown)


mail/auth.py
```diff
@@ -21,3 +25,57 @@ def authenticate(self, connection: poplib.POP3_SSL):
 
     def __eq__(self, other: Authenticator):
         return self.user == other.user and self.password == other.password
+
+
+class ModernAuthentication:
```
currycoder, line 30: Out of interest, how have you been testing this end to end? I assume that you have test creds and point to an integration env etc?  Could it be worth documenting that somewhere?

currycoder: And I take it that your testing confirmed you could read emails from the mailbox?

mail/tests/test_auth.py
```diff
@@ -27,3 +28,141 @@ def test_not_equal(self):
         equal_auth = BasicAuthentication("diff_user", "diff_password")
 
         self.assertNotEqual(auth, equal_auth)
+
+
+class ModernAuthenticationTests(SimpleTestCase):
+    def test_authenticates_connection_with_silent_acquisition(self):
```
currycoder, line 34: I don't really know the difference between "silent acquisition" and non-silent...  Could some docstrings be added to these tests to help explain?

mail/auth.py
kevincarrogan: Yes. This has all been tested with real credentials, however that whole thing is setup and dealt with by SRE and I've just been handed those credentials.

kevincarrogan: They will eventually end up in vault and we'll pull from there when I tie this all up with the rest of the code.

currycoder (approved):

PR #159, opened 2022-08-23
ICMSLST-1658 Sent EORI to CHIEF in correct field (RPA Trader ID).

MattHolmes123:


mail/icms_serializers.py
```diff
@@ -36,14 +36,32 @@ class AddressSerializer(serializers.Serializer):
...
     end_date = serializers.DateField(required=False, allow_null=True)
 
+    def validate_eori_number(self, value: str) -> str:
+        """Basic validation for EORI number.
+
+        https://www.tax.service.gov.uk/check-eori-number
+        """
+
+        # This may need to be extended to include other prefixes (XI) in the future.
+        if not value.upper().startswith("GB"):
```
saruniitr, line 54: For Northern Ireland it starts with XI - not sure if you need to support NI.

MattHolmes123: Not currently.
I spoke to several people at HMRC about this.
It is expected that importers supply the main GB EORI (this may change in the future but not right now)

saruniitr (approved):

PR #160, opened 2022-08-30
Upgrade python buildpack

kevincarrogan:
When deploying the latest changes for https://github.com/uktrade/lite-hmrc/pull/158 there was a failure when installing the new requirements (sub-dependencies from `msal`, more specifically `cryptography`).

Updating the buildpack has rectified the issues with installing the dependencies.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,7 +1,5 @@
-
----
 applications:
-  - name: lite-hmrc-dev
-    buildpack: https://github.com/cloudfoundry/python-buildpack.git#v1.6.30
+  - buildpacks:
+      - python_buildpack
     memory: 256M
     disk_quota: 512M


manifest.yml
```diff
@@ -1,7 +1,5 @@
-
----
 applications:
-  - name: lite-hmrc-dev
-    buildpack: https://github.com/cloudfoundry/python-buildpack.git#v1.6.30
+  - buildpacks:
+      - python_buildpack
```
currycoder, line 3: Seems we are going from having a version locked buildpack to "latest".  Is that OK?

kevincarrogan: Yep. That’s actually what I’ve been told to do in the past and having a locked buildpack is actually the wrong thing.

currycoder (approved):

PR #161, opened 2022-08-31
Ltd 2715 implement modern auth

kevincarrogan:
This switches our pop3 connections to use MS modern authentication.


local.env
```diff
@@ -50,3 +50,7 @@ SPIRE_STANDIN_EMAIL_PASSWORD=password
 SPIRE_STANDIN_EMAIL_POP3_PORT=995
 
 LITE_API_URL=https://lite.example.com
+
+AZURE_AUTH_CLIENT_ID=<FROM_VAULT>
```
currycoder, line 54: Is there a clear set of non-prod keys for these somewhere?  Is it worth a comment on how to find them?

mail/tests/test_select_email_for_sending.py
```diff
@@ -361,9 +393,17 @@ def test_case3_sending_of_pending_licencedata_and_usagedata_mails_2(self, email_
         send_mail.assert_called_once()
         self.assertEqual(mail.status, ReceptionStatusEnum.REPLY_PENDING)
 
+    @mock.patch("mail.libraries.routing_controller.get_spire_to_dit_mailserver")
+    @mock.patch("mail.libraries.routing_controller.get_hmrc_to_dit_mailserver")
```
currycoder, line 397: Sooo these tests were connecting to a real mailbox before..?

currycoder (approved):

local.env
kevincarrogan: I've added a link in the README to vault.

There will be non-production credentials in there as well as the ones for the production application.

mail/tests/test_select_email_for_sending.py
kevincarrogan: No. They weren't connecting to the real mailboxes, there are other functions mocked that was stopping that from happening.

What's changed here is that the authentication mechanism is now trying to connect to the real OAuth provider when it's being setup and that's what I'm getting around (with the smallest change I could do to achieve that).

PR #165, opened 2022-09-27
ICMSLST-1748 Add ICMS task to process licence reply emails sent from CHIEF.

MattHolmes123:


mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+    """Downloads licenceReply and usageData emails from HMRC mailbox and stores in Mail model."""
+    logger.info("Checking for reply and usage emails")
+
+    auth = _get_hmrc_mailbox_auth()
+    mailbox_hostname = settings.INCOMING_EMAIL_HOSTNAME
+    port = settings.INCOMING_EMAIL_POP3_PORT
+
+    with transaction.atomic(), pop3.get_connection(auth, mailbox_hostname, port) as con:
+        try:
+            for msg_id in pop3.list_messages_ids(con):
```
saruniitr, line 28: You are reading a message and also processing in the same function. If there any issue while processing eg in `_save_licence_reply_email(mail)` then it doesn't get deleted and it is partially processed. In the next iteration it gets processed again.
I see that it is in an atomic block but I think it is better to separate it out.

Also this assumes all the messages are genuine and coming from expected source.
In case of `lite-hmrc` we look at message headers and ensure it is coming from the expected senders. I have also seen Microsoft newsletter emails show up in this mailbox which gets ignored in our case.

The message ids that `pop3.list_messages_ids()` return - are they unique?
We have seen these being reused, although we were using `list()` method to get all ids and extract them from there - hence we look at message headers and extract `message-id` from there which is a unique identifier.

Is there no need to maintain whether a message has been read/unread. In case of lite-hmrc there is a table that saves message-id and its read status so when pulling new emails their `ids` are checked against this to filter unread emails.

mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+#
+# def send_usage_data_to_icms():
+#     """Checks Mail model for any usage data records to send to ICMS."""
+#     raise NotImplementedError
+
+
+def _get_hmrc_mailbox_auth() -> Authenticator:
+    # TODO: Replace with ModernAuthentication when needed
+
+    return BasicAuthentication(
```
saruniitr, line 64: This is still using basic auth. You need to switch to modern auth https://github.com/uktrade/lite-hmrc/blob/master/mail/auth.py#L30 as MS would've disabled basic auth at the end of September.

mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+        user=settings.INCOMING_EMAIL_USER,
+        password=settings.INCOMING_EMAIL_PASSWORD,
+    )
+
+
+def _save_licence_reply_email(reply_email: email.message.EmailMessage) -> None:
+    """Save the incoming licence reply email to the database."""
+
+    subject = reply_email.get("Subject")
+    run_number = _get_run_number_from_subject(subject)
```
saruniitr, line 74: The format used in https://github.com/uktrade/lite-hmrc/blob/3b6a89383b8a27c4d82e43a575e2258cd75ea313/mail/icms/tasks.py#L104 is not correct.
It will be similar to https://github.com/uktrade/lite-hmrc/blob/3b6a89383b8a27c4d82e43a575e2258cd75ea313/mail/icms/tasks.py#L108

mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+        raise ValueError("Only one attachment is accepted per licence reply email.")
+
+    reply_file = attachments[0]
+    file_name = reply_file.get_filename()
+
+    # e.g. b"1\\fileHeader\\SPIRE\\CHIEF\\licenceData\\202203171606\\1\\N\r\n..."
+    payload_bytes = reply_file.get_payload(decode=True)
+
+    ld = LicenceData.objects.get(hmrc_run_number=run_number)
+    mail = Mail.objects.select_for_update().get(id=ld.mail.id)
```
saruniitr, line 88: I think before updating the details below I think you should whether the `mail` is in the correct status and raise if that is not the case.
In our case we noticed that sometimes we are losing mails sent to HMRC/SPIRE. As in the mail is sent from our side but not received by the recipient so we manually resend (after confirming by email etc). During this time we may get new emails and can affect the order. These are edge cases but can happen so we check before updating that is in correct expected status besides the run number.

mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+
+    mail.status = ReceptionStatusEnum.REPLY_RECEIVED
+    mail.response_filename = file_name
+    mail.response_data = payload_bytes.decode()
+    mail.response_date = timezone.now()
+    mail.response_subject = subject
+
+    mail.save()
+
+    logger.info("Updated mail instance: %s with licence reply", mail.id)
```
saruniitr, line 98: Including subject/filename in the log message helps I think as it will provide run number etc which is not obvious from `id`

mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _get_run_number_from_subject(s: str) -> int:
+    try:
+        if "licenceReply" in s:
+            # e.g. CHIEF_licenceReply_29229_202209201442 -> 29229
```
saruniitr, line 104: This is not the correct format, it will be similar to the one below at L108

mail/models.py
```diff
@@ -67,8 +75,9 @@ def save(self, *args, **kwargs):
 
         super(Mail, self).save(*args, **kwargs)
 
-        if self.response_data and ReplyStatusEnum.REJECTED in self.response_data:
-            self.notify_users(self.id, self.response_date)
+        if settings.SEND_REJECTED_EMAIL:
```
saruniitr, line 78: Do we need this to be configurable? trying to understand when we would want this to be turned off.

mail/icms/tasks.py
MattHolmes123: I am not really processing anything in this task, it just saves the email filename & attached file.

`process_licence_reply_and_usage_emails` does the following:
- reads the email
- saves the email payload to the db
- Schedules the email for deletion

If there is an exception the db transaction is rolled back and `con.rset()` is called to clear any scheduled email deletes.

Currently if there is an email with a subject it doesn’t understand then nothing is processed and an error is thrown (in future it may delete these unknown emails or move them to another mailbox folder).

I should check for the expected sender for message types it does understand so I will add that.

`pop3.list_messages_ids()` is using the standard library `poplib.list` function, it just parses the msg_id from the response.

I am not interested right now in maintaining a db table to see if a message has been read or not as I’m deleting emails as soon as I have stored them in the database.
The next task I write will check for emails in the db to parse the contents and send that information back to ICMS.
I considered using `imaplib` instead of pop3 so I could use soft deletes to move the read emails to another mailbox folder instead of what I have now.

mail/icms/tasks.py
MattHolmes123: The test mailbox I am using is still working using BasicAuth, as you can see I have a TOOD to change it. 
I will create a story to change it over and implement it before we go live / I am forced to by BasicAuth no longer working.

mail/icms/tasks.py
MattHolmes123: I have been looking at the emails IMCS V1 gets back from CHIEF and the format is correct for both licenceReply and usageData.

mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+    payload_bytes = reply_file.get_payload(decode=True)
+
+    ld = LicenceData.objects.get(hmrc_run_number=run_number)
+    mail = Mail.objects.select_for_update().get(id=ld.mail.id)
+
+    mail.status = ReceptionStatusEnum.REPLY_RECEIVED
+    mail.response_filename = file_name
+    mail.response_data = payload_bytes.decode()
+    mail.response_date = timezone.now()
+    mail.response_subject = subject
```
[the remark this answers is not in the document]
MattHolmes123: Yes I don’t think I need / want to do that that at this time.
The Mail record will stay as a licence_data extract_type.
Ideally a LicenceData class should have multiple linked Mails, one for each message type (licenceData / licenceReply).

mail/icms/tasks.py
MattHolmes123: Sure can do

mail/icms/tasks.py
MattHolmes123: Again these formats I added match up to the subjects I am seeing in the emails coming from CHIEF to ICMS V1.

mail/models.py
MattHolmes123: I made that configurable because right now there is no requirement for ICMS to send an email if any licences are rejected in the licenceReply file.
Because that code exists in the save method of the mail class it made it very hard for me to not call it (without making it configurable behind a setting).

mail/icms/tasks.py
saruniitr: :+1:, I guess it slightly different in ICMS then.

mail/icms/tasks.py
saruniitr: :+1:

mail/models.py
saruniitr: :+1:

mail/icms/tasks.py
saruniitr: :+1:

mail/icms/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+    auth = _get_hmrc_mailbox_auth()
+    mailbox_hostname = settings.INCOMING_EMAIL_HOSTNAME
+    port = settings.INCOMING_EMAIL_POP3_PORT
+
+    with transaction.atomic(), pop3.get_connection(auth, mailbox_hostname, port) as con:
+        try:
+            for msg_id in pop3.list_messages_ids(con):
+                logger.debug("Processing msg_id %s", msg_id)
+
+                mail = pop3.get_email(con, msg_id)
```
MattHolmes123, line 31: TODO: Check sender is correct.

mail/icms/tasks.py
MattHolmes123: Added TODO to change before going live.

saruniitr (approved):

kevincarrogan (approved):

PR #170, opened 2022-10-31
Add logging for failures to ingest licence data

kevincarrogan:
We have a licence that appears to be failing to be created, however we respond with a 400 and the sender doesn't seem to throw any logs to indicate when this has happened. We also don't get the full response back in our logs when we send this back so we don't really know what's causing this problem.

This adds logging so that we're aware of any failures when creating licence data.


depsiatwal (approved):
lgtm

PR #171, opened 2022-11-02
ICMSLST-1762 Add support for replace CHIEF messages for ICMS.

MattHolmes123:
Other changes:
  - Send the "lite_id" back to ICMS instead of the transaction_ref.
  - Add "licence_payloads" attribute to LicenceData class.
  - Fix failing healthcheck when running with ICMS config.


mail/models.py
```diff
@@ -165,14 +166,25 @@ class Meta:
...
+        if settings.CHIEF_SOURCE_SYSTEM == ChiefSystemEnum.SPIRE:
+            LicenceIdMapping.objects.get_or_create(lite_id=self.lite_id, reference=self.reference)
 
 
 class LicenceIdMapping(models.Model):
     lite_id = models.UUIDField(primary_key=True, null=False, blank=False)
     reference = models.CharField(null=False, blank=False, max_length=35, unique=True)
 
 
+class LicenceDataLicencePayload(models.Model):
```
saruniitr, line 181: Is this model required or `LicencePayload` cannot be used instead?

MattHolmes123: yes LicencePayload can be used.

It would be a case of adding this single field:
`payloads = models.ManyToManyField(LicencePayload, related_name="+")`

I didn't do it this way as I didn't want this change to impact on the lite workflow in any way.

it would be useful for you guys too though (to see LicencePayload linked to a LicenceData record).

It would be more useful than this column that is currently on the LicenceData model:
`licence_ids = models.TextField()`

MattHolmes123: It's up to you what you would prefer me to do

MattHolmes123: At the moment the model is only populated when running for ICMS:
https://github.com/uktrade/lite-hmrc/pull/171/files/b6d9b8910ab9999a4ca14f2a553bc8238a3fb714#diff-197ad7a93c2498fccf48c6a58238cfd63a388a5ef79bc7f884ee6c24ad77c730R223-R225

saruniitr: Yeah I think it will be helpful to see `LicencePayload` linked to a `LicenceData` record, I don't know why this wasn't linked in the first place. But linking it now means we need to make changes in `lite-hmrc` and probably in `lite-api` also which is additional work.

MattHolmes123: If I make the suggested change you wouldn't need to do anything straight away.
It would simply exist as a reference.

ICMS uses this link to send the correct UUID back to ICMS.

saruniitr (approved):

PR #176, opened 2022-12-07
ICMSLST-1797 Add celery and ICMS tasks.

MattHolmes123:
Changes:

- Add celery for running background tasks.
- Add flower to monitor tasks locally.
- Add celery beat config for running all icms tasks.
- Refactor send_licence_data_to_hmrc to use shared function (send_licence_data_to_hmrc_shared)
- Update Procfile to include celery processes
- Fixed a mohawk bug when running lite-hmrc in cloud foundry.

Example of flower when running locally:
image

The commands I ran to configure redis for icms-hmrc in cloud foundry
- cf create-service redis tiny-ha-5.x icms-hmrc-dev-redis
- cf bind-service icms-hmrc-dev icms-hmrc-dev-redis
- cf scale icms-hmrc-dev --process celery_beat -i 1
- cf scale icms-hmrc-dev --process celery_worker -i 1
- cf scale icms-hmrc-dev --process dbt_worker -i 1
- cf scale icms-hmrc-dev --process worker -i 0

image

New environment variables added:
- ICMS_API_URL
- APP_ENV

**APP_ENV will need setting in vault for lite-hmrc**


conf/authentication.py
```diff
@@ -46,12 +47,21 @@ def _authenticate(request):
...
 
+    if settings.CHIEF_SOURCE_SYSTEM == ChiefSystemEnum.ICMS:
+        url = request.build_absolute_uri()  # pragma: no cover
+    else:
+        # I believe this is a bug, but I will leave to the lite team to investigate.
+        # e.g. "https://".replace("http", "https") == "httpss://"
+        # Original comment below:
+
+        # build_absolute_uri() returns 'http' which is incorrect since our clients communicate via https
+        url = request.build_absolute_uri().replace("http", "https")
```
MattHolmes123, line 58: The change I have added works for http & https when running with ICMS (docker & cloud foundry).
I am not sure what lite mohawk senders are doing but I have left the behaviour as it was before in your environments.
You should look in to this though because this line is a bug!

MattHolmes123: Django docs even have a not about letting webservers redirecting http to https instead of application code:
https://docs.djangoproject.com/en/4.1/ref/request-response/#django.http.HttpRequest.build_absolute_uri

build_absolute_uri always uses the same scheme that the current request has

PR #179, opened 2022-12-16
LTD-3148-fix-trader-address-multiline 

depsiatwal:
Initial fix was to also sanitise  the trade address key under [“organisation”] however this fix failed since more than 1 line was going over the 35 charter limit. Since a trader address is split over 5 lines. We have now created a new function which combines the 5 lines and then splits into 35 chars this has a couple of implications :

Trader address format gets changes 

We truncate if we exceed 5 lines

For the later we log in sentry this is the safe fix for now . Long term we should restrict UI from going over.

### DIFF
--- a/mail/libraries/lite_to_edifact_converter.py
+++ b/mail/libraries/lite_to_edifact_converter.py
@@ -248,7 +248,7 @@ def sanitize_foreign_trader_address(trader):
 
 def sanitize_trader_address(trader):
     """
-    Foreign trader address is split into 5 lines.
+    Trader address is split into 5 lines.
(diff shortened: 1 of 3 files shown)


mail/libraries/lite_to_edifact_converter.py
```diff
@@ -246,6 +246,35 @@ def sanitize_foreign_trader_address(trader):
     return trader
 
 
+def sanitize_trader_address(trader):
+    """
+    Foreign trader address is split into 5 lines.
```
saruniitr, line 251: It is actually just trader here

saruniitr (approved):

PR #184, opened 2022-12-29
LTD-3175-ablity-skip-payload-object 

depsiatwal:
This allows payload object to be skipped 
a management command can be used to set the flag.


mail/management/commands/set_skip_flag_on_payload.py
```diff
@@ -0,0 +1,28 @@
...
+from mail.models import LicencePayload
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument("--reference", type=str, nargs="?", help="License Reference number of the license payload")
+        parser.add_argument(
+            "--skip_process", type=str, nargs="?", help="To skip processing set to true", default="False"
+        )
+        parser.add_argument("--dry_run", type=str, nargs="?", help="Is it a test run", default="True")
```
saruniitr, line 12: we can set the type to `bool` and then no need to compare with `false/true` below

mail/management/commands/set_skip_flag_on_payload.py
```diff
@@ -0,0 +1,28 @@
+from django.core.management.base import BaseCommand
+
+from mail.models import LicencePayload
+
+
+class Command(BaseCommand):
```
saruniitr, line 6: could you add help text with `help = `

mail/management/commands/set_skip_flag_on_payload.py
```diff
@@ -0,0 +1,28 @@
...
+            "--skip_process", type=str, nargs="?", help="To skip processing set to true", default="False"
+        )
+        parser.add_argument("--dry_run", type=str, nargs="?", help="Is it a test run", default="True")
+
+    def handle(self, *args, **options):
+        dry_run = options.pop("dry_run")
+        reference = options.pop("reference")
+        skip_process = options.pop("skip_process")
+
+        payload = LicencePayload.objects.get(reference=reference)
```
saruniitr, line 19: I think it is better to check if it is still not processed before marking it as skip

mail/management/commands/set_skip_flag_on_payload.py
```diff
@@ -0,0 +1,28 @@
...
+        payload.skip_process = skip_process
+
+        if dry_run.lower() == "false":
+            payload.save()
+            self.stdout.write(self.style.SUCCESS(f"Reference  {reference} skip_process set to {payload.skip_process}"))
```
saruniitr, line 24: using `logging` is better as we get that in ELK and also f-strings are not recommended, old-style formatting is preferred as that is lazy.

mail/models.py
```diff
@@ -158,6 +158,8 @@ class LicencePayload(models.Model):
     data = models.JSONField(default=dict)
     received_at = models.DateTimeField(default=timezone.now)
     is_processed = models.BooleanField(default=False)
+    # This allows us to skip License requests to be skipped
+    skip_process = models.BooleanField(default=False)
```
saruniitr, line 162: We also need to make below change,

```
commit 5f904a03bf6fd1a8b129d29b585fd07841bb084a
Author: Kevin Carrogan <kevin.carrogan@gmail.com>
Date:   Mon Jan 31 12:10:29 2022 +0000

    Add default value for skip field
    This should already be handled by the default value of this field in the model definition however this is raising errors with this field not being fully populated so adding this more explicit value to get around this
```

saruniitr (approved):

PR #191, opened 2023-04-13
Add command to find eori number

kevincarrogan:
We seem to be getting an increase in support requests to find out the EORI number for licences that we have sent to HMRC. This management command gets this information for us without having to poke around in the database manually.

### DIFF
--- a/mail/management/commands/find_eori_number.py
+++ b/mail/management/commands/find_eori_number.py
@@ -1,0 +1,86 @@
+from django.core.management import BaseCommand
+
+from mail.enums import ReceptionStatusEnum, ReplyStatusEnum
+from mail.models import Mail
+
+
+class LicenceNotFoundError(Exception):
+    pass
+
+
+class Command(BaseCommand):
+    help = """Given a licence number will find the EORI number that we sent to
+    HMRC
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            dest="licence_number",
+            type=str,
+            help="Licence number to find EORI number",
+        )
+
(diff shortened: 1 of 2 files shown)


mail/management/commands/find_eori_number.py
```diff
@@ -0,0 +1,86 @@
...
+                continue
+            licence_line = next(lines)
+            while licence_line:
+                tokens = licence_line.split("\\")
+                if tokens[1] != "trader":
+                    licence_line = next(lines)
+                    continue
+                if tokens[1] == "end":
+                    break
+                return tokens[3]
```
saruniitr, line 42: this should be `tokens[2]`

mail/management/commands/find_eori_number.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def get_trade_number(self, sent_data, licence_number):
+        lines = iter(sent_data.split("\n"))
+        line = next(lines)
+        while line:
+            formatted_licence_number = self.get_formatted_licence_number(licence_number)
+            if f"licence\\{formatted_licence_number}" not in line:
+                line = next(lines)
+                continue
+            licence_line = next(lines)
```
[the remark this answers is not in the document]
kevincarrogan: `licence_line` here just means it's a line that's part of the licence, not specifically that it's the trader line. The next loop just goes through to find the trader line.

mail/management/commands/find_eori_number.py
saruniitr: `tokens[3]` is correct, please ignore above comment.

PR #194, opened 2023-05-17
Add MIR to UnitMapping for lite serializer

hnryjmes:
This is a fix for an error observed in Sentry when creating license data from lite, as a good has `MIR` as the unit in the request body from lite-api, and this is not supported by the lite-hmrc lite good serializer. This changes the UnitMapping to include the `MIR` unit with the appropriate code as in the UK Trade Tariff list (https://www.gov.uk/government/publications/uk-trade-tariff-quantity-codes/uk-trade-tariff-quantity-codes)

### DIFF
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -193,6 +193,7 @@ class UnitMapping(enum.Enum):
     MTR = 57  # meters
     LTR = 94  # litre
     MTQ = 2  # meters_cubed
+    MIR = 74  # millilitre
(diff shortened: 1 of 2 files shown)


depsiatwal (commented):
no tests ?

saruniitr (approved):
we don't seem to have any tests that check for how the file is generated with different units

kevincarrogan (approved):

PR #196, opened 2023-06-13
 LTD-3804-limit-hmrc-emails 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-3804

We had authentication issue with MS Mail. 

The account appears fine now but we have no formal coms of why we were locked out 

One thing we can do to mitigate is to make the number of emails we pull configurable. we currently pull 500. 

We should default to 100 which will realistically reduce how many call we make. Currently this is made against two inboxes.


saruniitr (approved):

mail/libraries/mailbox_service.py
```diff
@@ -70,13 +70,17 @@ def get_message_iterator(pop3_connection: POP3_SSL, username: str) -> Iterator[T
...
+    # Check only the emails specified in the setting
     # Since we don't delete emails from these mailboxes the total number can be very high over a perio
     # and increases the processing time.
     # The mails is a list of message number and size - message number is an increasing value so the
     # latest emails will always be at the end.
-    mail_message_ids = [get_message_id(pop3_connection, m.decode(settings.DEFAULT_ENCODING)) for m in mails[-500:]]
+    mail_message_ids = [
+        get_message_id(pop3_connection, m.decode(settings.DEFAULT_ENCODING))
+        for m in mails[-incoming_email_check_limit:]
+    ]
```
antroy-madetech, line 83: How often do we poll for emails? Do we have a estimate for the number of new messages we get each time we poll?

Just wondering if there's a danger that (e.g.) 120 messages come through, we process the last 100, and 20 get left unread forever. Guessing the 500 was a large number for this reason. Perhaps would be safer to grab another batch if the read_messages below is empty (i.e. all messages are new)

antroy-madetech (approved):

PR #215, opened 2024-01-03
Upgrade to django 4.2

Tllew:
Upgrading django to 4.2


hnryjmes (approved):
pipfile related changes and import name changes look fine to me, not sure why the tests are failing but will leave you to figure that out

mail/migrations/0001_initial.py
```diff
@@ -85,20 +84,20 @@ class Migration(migrations.Migration):
             name="UsageUpdate",
             fields=[
                 ("id", models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
-                ("licence_ids", jsonfield.fields.JSONField(default=dict)),
+                ("licence_ids", models.JSONField(default=dict)),
```
saruniitr, line 87: I think we usually don't change migrations that are already applied

mail/models.py
```diff
@@ -125,8 +125,8 @@ class UsageData(models.Model):
     mail = models.ForeignKey(Mail, on_delete=models.DO_NOTHING, null=False)
     spire_run_number = models.IntegerField()
     hmrc_run_number = models.IntegerField()
-    has_lite_data = models.NullBooleanField(default=None)
-    has_spire_data = models.NullBooleanField(default=None)
+    has_lite_data = models.BooleanField(null=True)
+    has_spire_data = models.BooleanField(null=True)
```
saruniitr, line 129: Updated migration file is not committed

mail/migrations/0001_initial.py
Tllew: It errors without it, this was something I found in Gurdeep's lite-api changes too

mail/models.py
```diff
@@ -192,12 +193,14 @@ class OrganisationIdMapping(models.Model):
 
 
 class GoodIdMapping(models.Model):
+    id = models.AutoField(primary_key=True)
```
kevincarrogan, line 196: Instead of adding this for each model I think this can just be configured in one places in the AppConfig.

https://docs.djangoproject.com/en/5.0/ref/applications/#django.apps.AppConfig.default_auto_field

Tllew: Yeah this is definitely an option, wasn't sure if it as cleaner to go explicit though?

saruniitr (approved):

PR #216, opened 2024-01-09
Ltd 4585 add celery

Tllew:


Procfile
```diff
@@ -1,3 +1,4 @@
 # See gunicorn.conf.py for more configuration.
 web: python manage.py migrate && gunicorn conf.wsgi:application
 worker: python manage.py process_tasks --log-std
+celeryworker: celery -A conf worker -l info
```
kevincarrogan, line 4: Should we also have a celery beat worker here too?

docker-compose.yml
```diff
@@ -30,12 +32,61 @@ services:
...
+  celery:
+    build: .
+    volumes:
+      - .:/app
+    env_file: .env
+    links:
+      - lite-hmrc-postgres
+      - redis
+    depends_on:
+      - lite-hmrc-intg
```
kevincarrogan, line 57: It's interesting that we have the dependency this way around, I would imagine that we'd want to say that the main service depends on this being up.

docker-compose.yml
```diff
@@ -30,12 +32,61 @@ services:
...
+    expose:
+      - 6379
+    ports:
+      - 6379:6379
+    networks:
+      - lite
 
+networks:
+  lite:
+    external: true
```
kevincarrogan, line 90: I don't think we need to setup an external network here.

core/tests/test_celery_tasks.py
```diff
@@ -0,0 +1,29 @@
...
+from django.conf import settings
+from django.utils import timezone
+from rest_framework.test import APITestCase
+
+from core import celery_tasks
+from mail.enums import ReplyStatusEnum
+from mail.models import Mail
+
+
+class FlagsUpdateTest(APITestCase):
```
kevincarrogan, line 12: I think this test needs to be renamed.

conf/settings.py
```diff
@@ -255,3 +259,34 @@
 AZURE_AUTH_TENANT_ID = env.str("AZURE_AUTH_TENANT_ID")
 
 SEND_REJECTED_EMAIL = env.bool("SEND_REJECTED_EMAIL", default=True)
+
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
+
+if VCAP_SERVICES:
+    if "aws-s3-bucket" not in VCAP_SERVICES:
+        raise Exception("S3 Bucket not bound to environment")
```
kevincarrogan, line 267: I think you might have copied and pasted too much from the `lite-api` as we don't need to do the bucket check for this service.

kevincarrogan: It's actually probably a good idea to push this onto a test environment as I _think_ you would have seen this break quite quickly once you deploy it.

Procfile
currycoder: I think that would be handy while we are here

core/tests/test_celery_tasks.py
currycoder: 👀 https://github.com/uktrade/lite-api/pull/1050/files#diff-e2b82cb4633cc863bbb13ebd2337311d8818306814adf656cf6d8e3c157d6953R7

conf/settings.py
Tllew: Ahh oops, I thought I took all the aws-s3 references out of settings.py

docker-compose.yml
Tllew: I think this was something I added in the desperate attempts to get rid of an ampq:// connection interrupted error I was having, but I needed to `cp local.env .env`

Procfile
Tllew: This has now been added

conf/settings.py
```diff
@@ -255,3 +259,34 @@
...
+if REDIS_BASE_URL:
+    # Give celery tasks their own redis DB - future uses of redis should use a different DB
+    REDIS_CELERY_DB = env("REDIS_CELERY_DB", default=0)
+    is_redis_ssl = REDIS_BASE_URL.startswith("rediss://")
+    url_args = {"ssl_cert_reqs": "CERT_REQUIRED"} if is_redis_ssl else {}
+
+    CELERY_BROKER_URL = _build_redis_url(REDIS_BASE_URL, REDIS_CELERY_DB, **url_args)
+    CELERY_RESULT_BACKEND = CELERY_BROKER_URL
+
+CELERY_ALWAYS_EAGER = env.bool("CELERY_ALWAYS_EAGER", False)
```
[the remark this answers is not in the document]
Tllew: I've removed `CELERY_ALWAYS_EAGER` so we'll find out during testing

celerybeat-schedule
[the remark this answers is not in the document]
Tllew: Done

saruniitr (approved):
with the changes verified on an env, this looks good to me

currycoder (approved):

kevincarrogan (approved):

PR #218, opened 2024-01-11
Ltd-4583: Migrate the task that notifies rejected licences to use Celery

Tllew:
Replaced rejected mail with a celery task and implemeted celery into the task which previously was running without background_tasks, which I believe was an oversight

### DIFF
--- a/mail/celery_tasks.py
+++ b/mail/celery_tasks.py
@@ -1,0 +1,46 @@
+import logging
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+from celery import shared_task
+from django.conf import settings
+
+from mail.servers import smtp_send
+
+logger = logging.getLogger(__name__)
+
+MAX_ATTEMPTS = 3
+RETRY_BACKOFF = 180
+
+
+# Notify Users of Rejected Mail
+@shared_task(
+    autoretry_for=(Exception,),
+    max_retries=MAX_ATTEMPTS,
+    retry_backoff=RETRY_BACKOFF,
+)
+def notify_users_of_rejected_mail(mail_id, mail_response_date):
+    """If a rejected email is found, this task notifies users of the rejection"""
+
+    logger.info("Notifying users of rejected Mail [%s, %s]", mail_id, mail_response_date)
+
+    try:
+        multipart_msg = MIMEMultipart()
+        multipart_msg["From"] = settings.EMAIL_USER
+        multipart_msg["To"] = ",".join(settings.NOTIFY_USERS)
+        multipart_msg["Subject"] = "Mail rejected"
+        body = MIMEText(f"Mail [{mail_id}] received at [{mail_response_date}] was rejected")
+        multipart_msg.attach(body)
+
+        smtp_send(multipart_msg)
+    except Exception as exc:  # noqa
+        error_message = (
+            f"An unexpected error occurred when notifying users of rejected Mail "
+            f"[{mail_id}, {mail_response_date}] -> {type(exc).__name__}: {exc}"
+        )
+
+        # Raise an exception
+        # this will cause the task to be marked as 'Failed' and retried if there are retry attempts left
+        raise Exception(error_message)
+    else:
+        logger.info("Successfully notified users of rejected Mail [%s, %s]", mail_id, mail_response_date)
--- a/mail/celery_tasks.py
+++ b/mail/celery_tasks.py
@@ -1,13 +1,12 @@
-import logging
+from celery import shared_task
+from celery.utils.log import get_task_logger
+from django.conf import settings
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
 
-from celery import shared_task
-from django.conf import settings
-
 from mail.servers import smtp_send
 
-logger = logging.getLogger(__name__)
+logger = get_task_logger(__name__)
 
 MAX_ATTEMPTS = 3
 RETRY_BACKOFF = 180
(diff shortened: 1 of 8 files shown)


Tllew:
Tests all working locally, failing due to redis config which we're discussing at 11am

mail/celery_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+    try:
+        multipart_msg = MIMEMultipart()
+        multipart_msg["From"] = settings.EMAIL_USER
+        multipart_msg["To"] = ",".join(settings.NOTIFY_USERS)
+        multipart_msg["Subject"] = "Mail rejected"
+        body = MIMEText(f"Mail [{mail_id}] received at [{mail_response_date}] was rejected")
+        multipart_msg.attach(body)
+
+        smtp_send(multipart_msg)
+    except Exception as exc:  # noqa
```
stuaxo, line 36: Are there a narrower set of exceptions we can have here, just those that `.attach` and `smtp_send` raise ?

mail/celery_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+        error_message = (
+            f"An unexpected error occurred when notifying users of rejected Mail "
+            f"[{mail_id}, {mail_response_date}] -> {type(exc).__name__}: {exc}"
+        )
```
stuaxo, line 40: Worth using the celery logger here.

mail/tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def celery_sync(settings):
```
saruniitr, line 5: @currycoder taken from your PR.

If this is merged then there will be conflicts in your PR as `celery_tasks.py` is added in this also.

mail/celery_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+        smtp_send(multipart_msg)
+    except SMTPException as exc:  # noqa
+        error_message = (
+            f"An unexpected error occurred when notifying users of rejected licences "
+            f"[Id: {mail_id}, subject: {mail_response_subject}] -> {type(exc).__name__}: {exc}"
+        )
+
+        # Raise an exception
+        # this will cause the task to be marked as 'Failed' and retried if there are retry attempts left
+        raise exc(error_message)
```
kevincarrogan, line 44: I know that this is copied from the original but I do think that we should use this as an opportunity to fix this.

It's a bit odd to be taking the exception object and then calling that (because I don't think it will be the exception class here).

The pattern we've used elsewhere is to log the exception via either `log.error` or `log.exception` and then we can just re-raise the exception so I would recommend something like:

```
except SMTPException:
    logging.exception("An unexpected error occurred when notifying users of rejected licences [Id: {mail_id}, subject: {mail_response_subject}]")  # this will contain the exception information already so we don't need to encode it in the log message
    raise

File: mail/models.py [hunk 5]
```diff
@@ -93,10 +93,10 @@ def set_response_date_time(self, offset: int = 0):
         self.save()
 
     @staticmethod
-    def notify_users(id, response_date):
-        from mail.tasks import notify_users_of_rejected_mail
+    def notify_users(id, response_subject):
+        from mail.celery_tasks import notify_users_of_rejected_mail
```
kevincarrogan, line 97: Do we need to import this in the function or could we import this at the top of the file?

mail/celery_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+    try:
+        multipart_msg = MIMEMultipart()
+        multipart_msg["From"] = settings.EMAIL_USER
+        multipart_msg["To"] = ",".join(settings.NOTIFY_USERS)
+        multipart_msg["Subject"] = "Mail rejected"
+        body = MIMEText(f"Mail (Id: {mail_id}) with subject {mail_response_subject} has rejected licences")
+        multipart_msg.attach(body)
+
+        smtp_send(multipart_msg)
+    except SMTPException as exc:  # noqa
```
depsiatwal, line 36: This will only retry for  SMTPException is that the intended behaviour ?
What if DB is down or some other issue do we not want to retry

mail/test_celery_task.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class NotifyUsersOfRejectedMailTests(TestCase):
+    @mock.patch("mail.celery_tasks.smtp_send")
+    def test_send_success(self, mock_send):
+        settings = {
+            "EMAIL_USER": "test@example.com",  # /PS-IGNORE
+            "NOTIFY_USERS": ["notify@example.com"],  # /PS-IGNORE
+        }
+        with self.settings(**settings):
```
kevincarrogan, line 16: Is there really a need to instantiate this as a dictionary and then splat it into `settings`?

saruniitr: I actually forgot to fix this, will update.

mail/models.py
```diff
@@ -78,7 +78,7 @@ def save(self, *args, **kwargs):
 
         if settings.SEND_REJECTED_EMAIL:
             if self.response_data and ReplyStatusEnum.REJECTED in self.response_data:
-                self.notify_users(self.id, self.response_date)
+                self.notify_users(self.id, self.response_subject)
```
depsiatwal, line 81: why the change from response_data to response_subject ?

mail/celery_tasks.py
saruniitr: Yeah only need to retry if we fail to send email

mail/models.py
saruniitr: Subject contains more details than the date alone.
Subject is nothing but the file name which contains run number, timestamp etc.

kevincarrogan (approved):

saruniitr:
Verified on `demo` and it is sending email if a licence rejected response is received.

PR #219, opened 2024-01-12
[LTD-4581] Replace lite licence usage reporting task with celery equivalent

currycoder:
Replace the licence usage reporting task with a celery equivalent.

The task is kicked off when a new licence usage email comes in from HMRC; if the email is identified as containing licence usage info for a licence issued by lite, the task is executed which sends the information over to lite API.
This has been successfully tested on the UAT environment; by sending a sample usage data email to this environment and observing that the new celery task executes successfully.


mail/celery_tasks.py
```diff
@@ -1,14 +1,76 @@
...
+    if not lite_usage_data.has_spire_data:
+        lite_usage_data.mail.status = ReceptionStatusEnum.REPLY_RECEIVED
+        lite_usage_data.mail.save()
+
+    lite_usage_data.save()
+
+
+def _handle_exception(message, lite_usage_data_id):
+    error_message = f"Failed to send LITE UsageData [{lite_usage_data_id}] to LITE API -> {message} "
+    raise Exception(error_message)
```
kevincarrogan, line 71: I don't like how this, potentially, scrubs out the original exception that was raised. I know it's mostly replicating what was there before but I wonder if it's an opportunity to make this a bit better.

mail/celery_tasks.py
```diff
@@ -43,3 +105,73 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
...
+    logger.info("Preparing LITE UsageData [%s] for LITE API", lite_usage_data_id)
+
+    try:
+        lite_usage_data = UsageData.objects.get(id=lite_usage_data_id)
+        licences = UsageData.licence_ids
+    except UsageData.DoesNotExist:  # noqa
+        _handle_exception(
+            f"LITE UsageData [{lite_usage_data_id}] does not exist.",
+            lite_usage_data_id,
+        )
```
kevincarrogan, line 127: On first read, because I didn't know that `_handle_exception` itself raises a new exception, it wasn't obvious that execution was stopped here.

mail/celery_tasks.py
```diff
@@ -43,3 +105,73 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
...
+)
+def send_licence_usage_figures_to_lite_api(lite_usage_data_id):
+    """Sends HMRC Usage figure updates to LITE"""
+
+    logger.info("Preparing LITE UsageData [%s] for LITE API", lite_usage_data_id)
+
+    try:
+        lite_usage_data = UsageData.objects.get(id=lite_usage_data_id)
+        licences = UsageData.licence_ids
+    except UsageData.DoesNotExist:  # noqa
```
kevincarrogan, line 123: What is this `# noqa` for?

depsiatwal (commented):
LGTM

Appears we need more code cov though

mail/tests/test_send_licence_usage_figures_to_lite_api.py
```diff
@@ -149,38 +148,7 @@ def test_schedule_usages_for_lite_api_400_bad_request(self, put_request):
         self.usage_data.refresh_from_db()
         self.assertIsNone(self.usage_data.lite_sent_at)
 
-    @mock.patch("mail.tasks.schedule_max_tried_task_as_new_task")
-    @mock.patch("mail.tasks.Task.objects.get")
-    @mock.patch("mail.tasks.mail_requests.put")
-    def test_schedule_usages_for_lite_api_max_tried_task(self, put_request, get_task, schedule_new_task):
```
kevincarrogan, line 155: I presume this is going because we're trusting that celery does its retry logic and we don't need to test it?

mail/tasks.py
```diff
@@ -1,203 +1,26 @@
...
-    else:
-        # Get the task's current attempt number by retrieving the previous attempts and adding 1
-        current_attempt = task.attempts + 1
-
-        # Schedule a new task if the current task has been attempted MAX_ATTEMPTS times;
-        # HMRC Integration tasks need to be resilient and keep retrying post-failure indefinitely.
-        # This logic will make MAX_ATTEMPTS attempts to send licence changes according to the Django Background Task
-        # Runner scheduling, then wait TASK_BACK_OFF seconds before starting the process again.
-        if current_attempt >= settings.MAX_ATTEMPTS:
-            schedule_max_tried_task_as_new_task(lite_usage_data_id)
```
kevincarrogan, line 194: This is slightly different now if I'm reading this correctly.

In the new code we'll keep trying for `MAX_ATTEMPTS` as that's what we're telling celery to do, but this seems to originally be trying for `MAX_ATTEMPTS`, stopping and then just trying again from the beginning.

It seems odd but it looks like we may have now changed it so that we don't retry indefinitely.

currycoder: TBH I think this is a positive change.  Retrying forever just isn't great.  I think the most likely scenario would be that for whatever reason lite-api was down/inaccessible and this would previously just try/try/try again until a 200 response.

What we now have is that this will try 3 times (with a backoff) and then stop.  Practically though, we do have a backstop here; https://github.com/uktrade/lite-hmrc/pull/219/files/cb941d8b3bb66996f7195c7e28f2a0d3776e77e3#diff-5f1ac6ff7b5a7a5cca14413fe6825d4bdf1b07fb8ee45291e2b54aadd90dbb71R36 - when django comes up, any `UsageData` records that have not been sent to lite-api will result in a celery task being spawned.

What we could do is define a `on_failure()` method for the task in question and log a message that tells us some manual intervention is required to ensure that UsageData is delivered to lite in a timely fashion.  What do you think?

mail/celery_tasks.py
currycoder: This was a `git mv` with extra steps, so don't know.  I can remove it though.

mail/celery_tasks.py
```diff
@@ -43,3 +105,73 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
         raise
 
     logger.info("Successfully notified users of rejected licences found in mail with subject %s", mail_response_subject)
+
+
+@shared_task(
+    autoretry_for=(Exception,),
+    max_retries=MAX_ATTEMPTS,
```
hnryjmes, line 112: the diff is showing me that the `MAX_ATTEMPTS` variable this is using is the one defined above, which is unchanged from the original code

I'm a bit confused now because in the deleted code it shows `settings.MAX_ATTEMPTS` and so I'm wondering if this should be using an env variable for max attempts as well? or was this a conscious decision to hardcode it as 3 for the rewrite in celery?

it seems like the kind of thing it would be useful to have as an env variable so you can tune it to the magic number you think works. although it sounds like you've already tested this on an environment and possibly decided that already.

mail/celery_tasks.py
currycoder: Gone with this approach - how does that seem?

mail/tasks.py
currycoder: I've gone ahead with the suggestion I made at the end there ^

mail/celery_tasks.py
currycoder: We seem to be using 3 across the board everywhere for this celery porting work.  I think 3 is fine for this task as it probably gives enough time (with the backoff) for a network hiccup/unexpected app recycle/whatever to resolve.
As for whether this should be tunable as an env variable, I don't see why we would want to have this vastly different across different environments.

In short I'm happy with this as-is, but happy to discuss further if you have a recommendation.

mail/celery_tasks.py
```diff
@@ -1,14 +1,75 @@
...
+
+    if not lite_usage_data.has_spire_data:
+        lite_usage_data.mail.status = ReceptionStatusEnum.REPLY_RECEIVED
+        lite_usage_data.mail.save()
+
+    lite_usage_data.save()
+
+
+def _log_error(message, lite_usage_data_id):
+    logger.error("Failed to send LITE UsageData [{%s}] to LITE API -> {%s}", lite_usage_data_id, message)
```
saruniitr, line 70: I understand this was in existing but could this be removed and use the logger methods directly?

mail/celery_tasks.py
```diff
@@ -43,3 +104,92 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
...
+        except Exception as exc:  # noqa
+            _log_error(
+                f"An unexpected error occurred when parsing the response for LITE UsageData "
+                f"[{lite_usage_data_id}] -> {type(exc).__name__}: {exc}",
+                lite_usage_data_id,
+            )
+            raise
+        save_response(lite_usage_data, accepted_licences, rejected_licences, response)
+
+    logger.info("Successfully sent LITE UsageData [%s] to LITE API", lite_usage_data_id)
```
saruniitr, line 195: Similar to the log at https://github.com/uktrade/lite-hmrc/pull/219/files#diff-63432b460fa40fc4046935520c31567757e3a9793f7cc662a98862f4dd8d0019R155, could you add `licences` also to it?

mail/celery_tasks.py
```diff
@@ -43,3 +104,92 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
...
 
     logger.info("Successfully notified users of rejected licences found in mail with subject %s", mail_response_subject)
+
+
+class BaseTask(Task):
+    def on_failure(self, exc, task_id, args, kwargs, einfo):
+        message = (
+            """
+        Maximum attempts for send_licence_usage_figures_to_lite_api task exceeded - the task has failed and needs manual inspection.
+        See mail.apps.MailConfig for a potential retry mechanism.
```
saruniitr, line 114: As this can be used on other tasks, could we get task name to be able to identify the task that is failing?
Is `task_id` the same as task name?

hnryjmes (approved):

mail/celery_tasks.py
currycoder: Just making this so that the name of the base task class is more specific to this usage data task.

mail/celery_tasks.py
currycoder: Personally, I still like there's some value to having this consistent format for error messages.  Is this feedback critical?

saruniitr: No, can be left as is.
The `message` itself is descriptive enough and also `lite_usage_data_id` gets duplicated in some cases.

mail/celery_tasks.py
saruniitr: ok, I think it was `BaseTask` before which you renamed to be specific to usage data base task so this is fine now.

saruniitr (approved):

PR #221, opened 2024-01-18
Add a mail resender management command

currycoder:
Add a django management command which allows re-sending of lite-hmrc EDI data to a specific lite-hmrc environment.  This has been useful for the testing of the reworked licence usage task;  https://github.com/uktrade/lite-hmrc/pull/219


mail/management/commands/resend_incoming_email.py
```diff
@@ -0,0 +1,77 @@
...
+    environment.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--edi_file_path",
+            type=str,
+            nargs="?",
+            help="Path to local EDI file that should be sent. Contents can be sourced from a valid Mail ORM record on the lite-hmrc environment; "
+            "Mail.edi_data.",
```
saruniitr, line 22: Incoming from HMRC is mainly licence_reply and usage_data.
Is this command for both extract types?

currycoder: Yes, I don't see why we couldn't use it for both types.  I am assuming that the overall shape of the email is the same aside from the EDI data payload?

saruniitr: yeah payload differs and the spec is different for each type.
I was just wondering if it is verified that it works for both as you were mainly dealing with usage data.
It is not an issue if it doesn't support reply yet, we can add any missing support as we go.
Please add it to help text whichever the case it is.

saruniitr (approved):

saruniitr:
Could you change it so that we only run this when DEBUG is enabled (to prevent from running on on `prod`)?

hnryjmes:
I think this might get confused with the `resend_email` task, maybe not by us but if someone uses this in the future who is not so familiar with lite-hmrc.

maybe you could change the name of the file to have the `dev_` prefix if you want, there are others which are named this way. or you might have another suggestion of how to avoid this getting confused with the `resend_email` task.

I appreciate that you've documented it and there's help text but I feel like some people don't read that so we should try and anticipate that risk as well :)

currycoder:
> I think this might get confused with the `resend_email` task, maybe not by us but if someone uses this in the future who is not so familiar with lite-hmrc.
> 
> maybe you could change the name of the file to have the `dev_` prefix if you want, there are others which are named this way. or you might have another suggestion of how to avoid this getting confused with the `resend_email` task.
> 
> I appreciate that you've documented it and there's help text but I feel like some people don't read that so we should try and anticipate that risk as well :)

@hnryjmes I'm not a fan of the `dev_` prefix because I don't really feel like there's a clearly defined delineation on these tasks right now.  I've renamed the task to be `resend_edi_data_email` to try avoid the ambiguity of mixing it up with `resend_email`.

hnryjmes (approved):

PR #223, opened 2024-01-19
Migrate task that sends LITE licence details to HMRC to use Celery

saruniitr:
## Change description

Migrate the task that sends licence details to HMRC to schedule via Celery.  
This is a periodic task that runs every 10 minutes and checks if any licence payload available that needs sending to HMRC. If available then an email is constructed with those details and sent to HMRC.

[LTD-4582](https://uktrade.atlassian.net/browse/LTD-4582)

[LTD-4582]: https://uktrade.atlassian.net/browse/LTD-4582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mail/libraries/routing_controller.py
```diff
@@ -176,7 +176,7 @@ def send(email_message_dto: EmailMessageDto):
 
 
 def _collect_and_send(mail: Mail):
-    from mail.tasks import send_licence_data_to_hmrc
+    from mail.celery_tasks import send_licence_details_to_hmrc
```
saruniitr, line 179: there is a circular dependency in this fail hence it is retained in its current position

mail/models.py
```diff
@@ -95,6 +94,8 @@ def set_response_date_time(self, offset: int = 0):
 
     @staticmethod
     def notify_users(id, response_subject):
+        from mail.celery_tasks import notify_users_of_rejected_licences
+
```
saruniitr, line 98: there is a circular dependency hence it is reverted back to its previous position

healthcheck/tests/test_views.py
```diff
@@ -53,10 +55,7 @@ def test_healthcheck_service_unavailable_inbox_task_not_responsive(self):
         self.assertEqual(response.context["status"], status.HTTP_503_SERVICE_UNAVAILABLE)
 
     def test_healthcheck_service_unavailable_licence_update_task_not_responsive(self):
-        run_at = timezone.now() + timedelta(minutes=settings.LITE_LICENCE_DATA_POLL_INTERVAL)
-        task, _ = Task.objects.get_or_create(queue=LICENCE_DATA_TASK_QUEUE)
-        task.run_at = run_at
-        task.save()
+        current_app.tasks.pop(CELERY_SEND_LICENCE_UPDATES_TASK_NAME)
```
kevincarrogan, line 58: You can use `del` instead of `pop` if you're not using the returned value.

Why are we removing this though?

saruniitr: To make the healthcheck fail.
To make this check fail https://github.com/uktrade/lite-hmrc/blob/master/healthcheck/views.py#L50 I am deleting the task.
Previously with background tasks we were changing the `run_at` field.

mail/celery_tasks.py
```diff
@@ -43,3 +52,57 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
...
+            logger.info("Licence references [%s] marked as processed", licence_references)
+
+    except (EdifactValidationError, SMTPException):  # noqa
+        logger.exception("An unexpected error occurred when sending LITE licence updates to HMRC -> %s")
+        raise
+
+    logger.info(
+        "Successfully sent LITE issued licence (%s) updates in mail (%s) to HMRC", licence_references, mail.edi_filename
+    )
+    return True
```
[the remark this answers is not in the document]
saruniitr: Actually no, this is only for this call here,
https://github.com/uktrade/lite-hmrc/blob/master/mail/views.py#L98

This endpoint is used in one of the E2E test so added this to retain the current behaviour.

mail/libraries/routing_controller.py
```diff
@@ -204,7 +204,7 @@ def _collect_and_send(mail: Mail):
         if message_to_send_dto.receiver == SPIRE_ADDRESS and mail.extract_type == ExtractTypeEnum.LICENCE_DATA:
             # Pick up any LITE licence updates once we send a licence update reply email to SPIRE
             # so LITE does not get locked out of the queue by SPIRE
-            send_licence_data_to_hmrc(schedule=0)  # noqa
+            send_licence_details_to_hmrc.delay()  # noqa
```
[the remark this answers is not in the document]
saruniitr: We can actually remove this but potentially sending of lite licence details can get delayed without this as lite traffic is much less than spire.
As we know we can only send one message at a time until it is fully processed (i.e., status updated as reply_sent).
possible scenario,
1. we are waiting for a reply for spire message (status = reply_pending)
2. Licence issued in lite but we cannot send this just yet
3. reply comes and we relay back to spire. At this point we can send lite details but the task may not ready to run yet.
4. We get another message from spire and say `manage_inbox()` task managed to run so this will end up sending spire message and we again wait for reply (back to first step)

Because of this change we would've picked lite details at the end of 3 and sent to hmrc.

I think we can delete this. The delay won't be long, we may miss one two cycles but it will get picked up soon.

Also if we choose to retain then no issue if two tasks end up running because we acquire lock at the beginning. Second task ends up waiting anyway. And when it runs it sees no unprocessed payloads (assuming first one successful)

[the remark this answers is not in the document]
hnryjmes: > lite traffic is much less than spire

I think as long as the thing we are building allows for flexibility - what we do now is only so important, it's being able to change these tasks and schedules easily in the future that I think is important

hnryjmes (approved):

mail/libraries/routing_controller.py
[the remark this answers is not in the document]
saruniitr: Removing this condition as discussed.
Please see commit message for more details where the above is documented.

healthcheck/tests/test_views.py
stuaxo: Is current_app.tasks going to have the expected value for a unit test that runs after this ?

I wonder if it would be safer to patch this just for the duration of the test(s) that need it ?

saruniitr: There is a fixture which populates this again and also all the tests passing so no issues,
https://github.com/uktrade/lite-hmrc/pull/223/files#diff-f025a113158bd2b86b5fe216a11923f1a069b43d09f309cd9c1af3f74b2aa5a2R8

healthcheck/checks.py
```diff
@@ -55,9 +57,9 @@ def is_licence_payloads_processing():
 
 
 def is_lite_licence_update_task_responsive():
-    dt = timezone.now() + datetime.timedelta(seconds=settings.LITE_LICENCE_DATA_POLL_INTERVAL)
+    licence_update_task = current_app.tasks.get(CELERY_SEND_LICENCE_UPDATES_TASK_NAME)
```
currycoder, line 60: Cross posted from Tomos' PR which leads on from this one..

What does current_app.tasks.get(CELERY_MANAGE_INBOX_TASK_NAME) return? I'm not sure this is a reliable alternative to what we had before; before we were checking that a task was scheduled to run at some reasonable time in the future.

My guess is that current_app.tasks is a sort of task manifest rather than a list of scheduled/currently running tasks.

saruniitr: updated, found a better of way checking this.

saruniitr: @Tllew you may need to do similar in your PR for migrating `manage_inbox()` task so please check if this helps.

depsiatwal (approved):
lgtm

healthcheck/tests/test_views.py
stuaxo: Using a fixture with yield could be an option to make the cleanup part automatic.

I wonder if it's worth considering using freeze time, if this is about how long a health check takes to run.

saruniitr:
This is also verified on an environment, task sending mails as expected.

mail/libraries/routing_controller.py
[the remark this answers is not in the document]
stuaxo: This sounds good + fairly stateless.   What happens if one of the payloads always fails, could we get tasks stacking up behind it ?

BTW - https://github.com/steinitzu/celery-singleton is something we may want to try in future on task like this.

healthcheck/tests/test_views.py
saruniitr: Their scope is restricted to function for the relevant tests so it is automatic.
Actually his is not about how long health check takes to run, this is to ensure whether this task is responsive (getting scheduled periodically as expected or not).

mail/libraries/routing_controller.py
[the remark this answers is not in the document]
saruniitr: If it is always failing then depending on the failure it most likely needs a code change either in API or in lite-hmrc.
Tasks won't be stacking, if payload fails then task exits with a failure and we will get other alerts if payloads not processed for a while (explained in commit message for this change).
Even if there are multiple tasks it is not a problem as each task needs to take a lock before it updates any state so they are mutually exclusive.

healthcheck/checks.py
```diff
@@ -55,9 +57,22 @@ def is_licence_payloads_processing():
...
+    if CELERY_SEND_LICENCE_UPDATES_TASK_NAME not in current_app.conf.beat_schedule:
+        return False
+
+    task_schedule = current_app.conf.beat_schedule[CELERY_SEND_LICENCE_UPDATES_TASK_NAME]["schedule"]
+
+    ready_to_run, next_time_to_check = task_schedule.is_due(timezone.localtime())
+    if ready_to_run:
+        return ready_to_run
+
+    return next_time_to_check < settings.LITE_LICENCE_DATA_POLL_INTERVAL
```
currycoder, line 75: Trying to understand; how will this healthcheck ever return a failure value?

saruniitr: Discussed internally and we agreed to remove this check completely.

currycoder (approved):

healthcheck/checks.py
saruniitr: @Tllew fyi

PR #224, opened 2024-01-22
LTD-4579-healthcheck

seijihg:
https://uktrade.atlassian.net/browse/LTD-4586

### DIFF
--- a/healthcheck/tests/test_checks.py
+++ b/healthcheck/tests/test_checks.py
@@ -1,0 +1,85 @@
+import logging
+import unittest
+import pytest
+from unittest.mock import Mock
+from healthcheck.exceptions import HealthCheckException
+from healthcheck.checks import celery_health_check
+from celery.exceptions import TaskRevokedError, TimeoutError
+
+
+@pytest.fixture
+def mock_settings():
+    return {
+        "HEALTHCHECK_CELERY_TIMEOUT": 3,
+        "HEALTHCHECK_CELERY_RESULT_TIMEOUT": 3,
+        "HEALTHCHECK_CELERY_QUEUE_TIMEOUT": 3,
+    }
+
+
+@pytest.fixture
+def mock_logger():
+    logger = logging.getLogger("health-check")
+    logger.error = Mock()
+    logger.exception = Mock()
+    return logger
+
+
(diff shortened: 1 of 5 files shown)


healthcheck/tests/test_checks.py
```diff
@@ -0,0 +1,90 @@
...
+        "HEALTHCHECK_CELERY_QUEUE_TIMEOUT": 3,
+    }
+
+
+@pytest.fixture
+def mock_logger():
+    logger = logging.getLogger("health-check")
+    logger.error = Mock()
+    logger.exception = Mock()
+    return logger
```
kevincarrogan, line 24: There is already a built-in logger fixture https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture.

healthcheck/tests/test_checks.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@pytest.fixture
+def add_mock():
+    return Mock()
```
kevincarrogan, line 29: I don't think this is a very useful fixture.

I _think_ we also might already have `mocker` installed so you could just use the `mocker` fixture.

healthcheck/tests/test_checks.py
```diff
@@ -0,0 +1,90 @@
...
+def test_add_task():
+    result = add(4, 5)
+    assert result == 9
+
+
+def test_celery_health_check_successful(mock_settings, mock_logger, add_mock):
+    add_mock.apply_async.return_value.result = 8
+
+    with unittest.mock.patch("healthcheck.checks.add", add_mock), unittest.mock.patch(
+        "django.conf.settings", mock_settings
```
kevincarrogan, line 41: To mock django settings you can use the `settings` fixture instead of having to do a `mock.patch`.

healthcheck/tests/test_checks.py
kevincarrogan, line 1: In this repo we use unittest style class based tests but this is using pytest style tests. We should be consistent across the repo.

hnryjmes:
I think this PR can be closed if it is not being used and we have https://github.com/uktrade/lite-hmrc/pull/226

PR #225, opened 2024-01-23
Ltd 4584 celery manage inbox task

Tllew:
[LTD-4584](https://uktrade.atlassian.net/browse/LTD-4584)

piggy backing on Arun's changes to avoid future conflicts

[LTD-4584]: https://uktrade.atlassian.net/browse/LTD-4584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mail/apps.py
```diff
@@ -12,25 +12,13 @@ def initialize_background_tasks(cls, **kwargs):
...
-        Task.objects.filter(queue=MANAGE_INBOX_TASK_QUEUE).delete()
         Task.objects.filter(queue=LICENCE_DATA_TASK_QUEUE).delete()
 
         if settings.BACKGROUND_TASK_ENABLED:
-            # Tasks common to ICMS and LITE/SPIRE
-            send_licence_data_to_hmrc(repeat=settings.LITE_LICENCE_DATA_POLL_INTERVAL, repeat_until=None)  # noqa
-
             # LITE/SPIRE Tasks
             if settings.CHIEF_SOURCE_SYSTEM == ChiefSystemEnum.SPIRE:
-                manage_inbox(repeat=settings.INBOX_POLL_INTERVAL, repeat_until=None)  # noqa
```
currycoder, line 32: If `settings.INBOX_POLL_INTERVAL` is being ignored, let's also remove it from the settings module itself.

conf/celery.py
```diff
@@ -16,3 +17,15 @@
...
 app.autodiscover_tasks(related_name="celery_tasks")
 app.autodiscover_tasks(["core"], related_name="celery_tasks")
+
+# Define any regular scheduled tasks
+app.conf.beat_schedule = {
+    "send licence details to hmrc, periodic task every 10min": {
+        "task": "mail.celery_tasks.send_licence_details_to_hmrc",
+        "schedule": crontab(minute="*/10"),
+    },
+    "manage inbox, periodic task every 5min": {
```
currycoder, line 27: I believe this task is run every 10 mins on UAT/prod as the settings are overridden in vault; would recommend setting that statically here.

healthcheck/checks.py
```diff
@@ -63,9 +62,9 @@ def is_lite_licence_update_task_responsive():
 
 
 def is_manage_inbox_task_responsive():
-    dt = timezone.now() + datetime.timedelta(seconds=settings.INBOX_POLL_INTERVAL)
+    manage_inbox_task = current_app.tasks.get(CELERY_MANAGE_INBOX_TASK_NAME)
```
currycoder, line 64: What does `current_app.tasks.get(CELERY_MANAGE_INBOX_TASK_NAME)` return?  I'm not sure this is a reliable alternative to what we had before; before we were checking that a task was scheduled to run at some reasonable time in the future.

My guess is that `current_app.tasks` is a sort of task manifest rather than a list of scheduled/currently running tasks.

currycoder: In Arun's PR, we came to the conclusion that actually there was another check present which was better.  Another healthcheck was checking the *outcome* of the task; that there existed no unprocessed LicencePayloads older than a certain threshold.  This can tell us that the task is working without having to zoom in to the guts of the task itself.  Is there a similar healthcheck for this?  To tell us that mail is being processed?  If not, can we make one?

saruniitr: we have one, https://github.com/uktrade/lite-hmrc/blob/master/healthcheck/checks.py#L63
If we haven't received reply for any mail within the time configured then this fails.

healthcheck/checks.py
```diff
@@ -54,10 +54,31 @@ def is_licence_payloads_processing():
...
+    ready_to_run, next_time_to_check = task_schedule.is_due(timezone.localtime())
+    if ready_to_run:
+        return ready_to_run
+
+    return next_time_to_check < poll_interval
 
-    return Task.objects.filter(queue=MANAGE_INBOX_TASK_QUEUE, run_at__lte=dt).exists()
+
+def is_lite_licence_update_task_responsive():
+    return is_periodic_task_responsive(CELERY_SEND_LICENCE_UPDATES_TASK_NAME, settings.LITE_LICENCE_DATA_POLL_INTERVAL)
```
saruniitr, line 77: I removed this check in my PR

healthcheck/checks.py
```diff
@@ -54,10 +54,31 @@ def is_licence_payloads_processing():
...
+
+
+def is_manage_inbox_task_responsive():
+    return is_periodic_task_responsive(CELERY_MANAGE_INBOX_TASK_NAME, settings.INBOX_POLL_INTERVAL)
```
saruniitr, line 81: This is also not required anymore as below check covers this.

If there are any pending mails for certain period then below check fails then we will get to know that `manage_inbox()` task is not getting scheduled or the reply itself is delayed from HMRC side.

Tllew: I did notice that, I'll remove it in this too if it's the preferred solution.

currycoder (approved):
Looks good - I think there's a couple of lines that codecov is complaining about though!

saruniitr (approved):

PR #226, opened 2024-01-24
LTD-4579-health-check

seijihg:
Adapt custom healthcheck to be used with the django-health-check library.

https://uktrade.atlassian.net/browse/LTD-4586


healthcheck/checks.py
```diff
@@ -6,75 +6,80 @@
...
+                logger.error(error_message)
+                self.add_error(HealthCheckException(error_message))
+                mailbox_results.append(False)
+            else:
+                mailbox_results.append(True)
+            finally:
+                mailserver.quit_pop3_connection()
+
+        if not all(mailbox_results):
+            raise HealthCheckException("One or more mailboxes failed to authenticate.")
```
saruniitr, line 43: Is adding the error not enough?

My understanding is either one of them is enough but not both, could you deploy this to an env and see the response from health check url?

healthcheck/checks.py
```diff
@@ -6,75 +6,80 @@
...
+
+        if unprocessed_payloads.exists():
+            raise HealthCheckException("There are unprocessed licence payloads.")
+
+
+class LiteLicenceUpdateTaskHealthCheck(BaseHealthCheckBackend):
+    def check_status(self):
+        dt = timezone.now() - datetime.timedelta(seconds=settings.LITE_LICENCE_DATA_POLL_INTERVAL)
+        if not Task.objects.filter(queue=LICENCE_DATA_TASK_QUEUE, run_at__lte=dt).exists():
+            raise HealthCheckException("Lite Licence Update Task is not responsive.")
```
saruniitr, line 64: We decided to remove this check as the above payload health check covers it.

healthcheck/tests/test_checks.py
```diff
@@ -0,0 +1,145 @@
...
+import uuid
+import datetime
+
+from parameterized import parameterized
+from django.test import TestCase
+from unittest.mock import patch
+from django.conf import settings
+from django.utils import timezone
+
+from healthcheck.checks import (
```
saruniitr, line 11: The file `healthcheck/tests/test_views.py` is deleted and those tests are added in this file because of change in file name?
This will corrupt git history. When you want to rename a file, instead of deleting old one and creating a new one with a different name you can use below which preserves history.
`git mv path_to_file_current_name path_to_file_new_name`

conf/urls.py
```diff
@@ -20,7 +20,7 @@
 urlpatterns = [
     path("admin/", admin.site.urls),
     path("mail/", include("mail.urls")),
-    path("healthcheck/", include("healthcheck.urls")),
+    path("healthcheck/", include("health_check.urls")),
```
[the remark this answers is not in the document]
seijihg: yea since it will cover other checks which are in line with lite-api:

    "health_check.db",
    "health_check.cache",
    "health_check.storage",
    "health_check.contrib.migrations",
    "health_check.contrib.celery",
    "health_check.contrib.celery_ping",

conf/settings.py
```diff
@@ -38,6 +38,13 @@
     "background_task",
     "mail.apps.MailConfig",
     "healthcheck",
+    "health_check",
```
hnryjmes, line 41: I feel like we need some more descriptive naming here. It seems like these two apps do different things but naming them something so similar will get confusing

healthcheck/apps.py
```diff
@@ -1,6 +1,22 @@
 from django.apps import AppConfig
 
+from health_check.plugins import plugin_dir
 
-class HealthcheckConfig(AppConfig):
-    default_auto_field = "django.db.models.BigAutoField"
+
+class MyAppConfig(AppConfig):
     name = "healthcheck"
```
hnryjmes, line 7: continuing from my comment above... if we can't change the name `health_check` because it is a part of the library `django-health-check` maybe this can be changed instead?

healthcheck/checks.py
seijihg: One is for detailed error is being added to the internal class list `self.add_error(HealthCheckException(error_message))`, and `raise HealthCheckException("One or more mailboxes failed to authenticate.") `is more for overall checks have failed.

healthcheck/tests/test_checks.py
seijihg: I think its just different now? one was testing the view now it is testing the checks since view is being handled by django healthcheck lib now. I guess it is doing kinda the same. lemme rename it instead then.

healthcheck/checks.py
currycoder: Amplifying Arun's earlier assertion I see
```
        # The test code goes here.
        # You can use `self.add_error` or
        # raise a `HealthCheckException`,
        # similar to Django's form validation.
``` 
in the django-health-check docs.  Personally, I think `self.add_error()` above is enough - it means we end up with one exception per failed mailbox.  You can then drop `mailbox_results` entirely and stop doing the work twice.

currycoder (commented):
Great work moving this over to django-health-check!  Just a couple extra points on top of the review comments already.

PR #229, opened 2024-01-31
LTD-4645-healthcheck-app-readd

seijihg:
a bit of cleaning up since few things were removed when merged.

### DIFF
--- a/healthcheck/apps.py
+++ b/healthcheck/apps.py
@@ -10,11 +10,9 @@ class MyAppConfig(AppConfig):
         from .checks import (
             MailboxAuthenticationHealthCheck,
             LicencePayloadsHealthCheck,
-            ManageInboxTaskHealthCheck,
             PendingMailHealthCheck,
         )
 
         plugin_dir.register(MailboxAuthenticationHealthCheck)
         plugin_dir.register(LicencePayloadsHealthCheck)
-        plugin_dir.register(ManageInboxTaskHealthCheck)
         plugin_dir.register(PendingMailHealthCheck)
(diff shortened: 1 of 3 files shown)


healthcheck/apps.py
```diff
@@ -10,11 +10,9 @@ def ready(self):
         from .checks import (
             MailboxAuthenticationHealthCheck,
             LicencePayloadsHealthCheck,
-            ManageInboxTaskHealthCheck,
             PendingMailHealthCheck,
         )
 
         plugin_dir.register(MailboxAuthenticationHealthCheck)
         plugin_dir.register(LicencePayloadsHealthCheck)
-        plugin_dir.register(ManageInboxTaskHealthCheck)
```
currycoder, line 19: Is the code for `ManageInboxTaskHealthCheck` still present?  Would suggest removing it if so

seijihg: ah no, it was removed when @Tllew merged his PR

Tllew: I'm not sure how you can check the manageinbox task as it's all converted to celery, but if it works then happy to keep it in

Tllew (approved):

healthcheck/apps.py
currycoder: We decided to drop the task, but a little confused that it's still referenced here

saruniitr: Yeah it is a bit confusing to see these references but the code is not there.

saruniitr (approved):
Suggest deploying to an env and checking before merging this.

seijihg:
> Suggest deploying to an env and checking before merging this.

done checked on UAT and DEMO both working fine

PR #230, opened 2024-02-06
LTD-4652-send-email-task

seijihg:
https://uktrade.atlassian.net/browse/LTD-4652

The reason I have put in `tasks.py` instead of `celery_tasks.py `is because there were tons of circular import errors and it would be kinda painful to resolve all of it.

AIming at where emails are being sent out where I replaced with new celery task `send_email_task()` where its using memcache_lock to send out email one by one.


stuaxo:
I know we called it celery_tasks, but tasks.py is the standard - probably too late to change this throughout, but there we go.

mail/tasks.py
```diff
@@ -15,3 +26,41 @@ def emit_test_file():
...
+    if "message" in kwargs:
+        multipart_msg = build_email_message(kwargs["message"])
+
+    if "mail_response_subject" in kwargs:
+        multipart_msg = build_email_rejected_licence_message(kwargs["mail_id"], kwargs["mail_response_subject"])
+
+    hexdigest = md5(str(kwargs["mail_id"]).encode("utf-8")).hexdigest()
+    lock_id = f"send_email-{hexdigest}"  # Construct a unique lock ID
+
+    with memcache_lock(lock_id, "email_send_operation") as acquired:
```
kevincarrogan, line 56: I may be understanding this incorrectly, but doesn't this just acquire a lock for just this email when actually we want to lock to make sure this only ever gets run once at at time?

mail/tasks.py
```diff
@@ -15,3 +26,41 @@ def emit_test_file():
...
+        if acquired:
+            logger.info("Lock acquired, sending email")
+            try:
+                smtp_send(multipart_msg)
+                logger.info("Successfully notified users.")
+            except SMTPException as e:
+                logger.error(f"Failed to send email: {e}")
+                raise
+        else:
+            logger.info("Another send_email_task is currently in progress")
```
kevincarrogan, line 66: Doesn't this mean that we just fail to send an email if another task is already running? I can't work out how we are retrying or waiting for the lock to be released?

mail/tasks.py
seijihg: I have changed to global lock it should work correctly now.

mail/celery_tasks.py
```diff
@@ -79,38 +77,6 @@ def _log_error(message, lite_usage_data_id):
 CELERY_MANAGE_INBOX_TASK_NAME = "mail.celery_tasks.manage_inbox"
 
 
-# Notify Users of Rejected Mail
-@shared_task(
-    autoretry_for=(SMTPException,),
-    max_retries=MAX_ATTEMPTS,
-    retry_backoff=RETRY_BACKOFF,
-)
-def notify_users_of_rejected_licences(mail_id, mail_response_subject):
```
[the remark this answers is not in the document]
saruniitr: It will be mainly https://github.com/uktrade/lite-hmrc/pull/230/files#diff-63432b460fa40fc4046935520c31567757e3a9793f7cc662a98862f4dd8d0019L101 that needs refactoring into other task but may need few other bits also to meet requirements of other tasks.

mail/models.py
```diff
@@ -94,9 +94,9 @@ def set_response_date_time(self, offset: int = 0):
 
     @staticmethod
     def notify_users(id, response_subject):
-        from mail.celery_tasks import notify_users_of_rejected_licences
+        from mail.tasks import send_email_task
 
-        notify_users_of_rejected_licences.delay(str(id), response_subject)
+        send_email_task.delay(mail_id=id, mail_response_subject=response_subject)
```
saruniitr, line 99: there is no need to change this.

mail/libraries/routing_controller.py
```diff
@@ -185,7 +186,7 @@ def _collect_and_send(mail: Mail):
 
     if message_to_send_dto:
         if message_to_send_dto.receiver != SourceEnum.LITE and message_to_send_dto.subject:
-            send(message_to_send_dto)
+            send_email_task(mail_id=mail.id, message=message_to_send_dto)
             update_mail(mail, message_to_send_dto)
```
saruniitr, line 190: Once the sending bit is refactored you may probably need to move the `update_mail()` also in that task as we can only do that once the email is sent successfully.

mail/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    status = cache.add(lock_id, "locked", LOCK_EXPIRE)
+    try:
+        yield status
+    finally:
+        if time.monotonic() < timeout_at and status:
+            cache.delete(lock_id)
+
+
+@shared_task(bind=True, autoretry_for=(SMTPException,), max_retries=MAX_ATTEMPTS, retry_backoff=RETRY_BACKOFF)
+def send_email_task(self, **kwargs):
```
saruniitr, line 31: I recommend this to be moved into `celery_tasks.py` itself.

We configured to discover all tasks from that file here,
https://github.com/uktrade/lite-hmrc/blob/master/conf/celery.py#L19

Keeping some of them in other files is not consistent.

mail/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    with memcache_lock(global_lock_id) as acquired:
+        if acquired:
+            logger.info("Global lock acquired, sending email")
+            try:
+                if "message" in kwargs:
+                    multipart_msg = build_email_message(kwargs["message"])
+                elif "mail_response_subject" in kwargs and "mail_id" in kwargs:
+                    multipart_msg = build_email_rejected_licence_message(
+                        kwargs["mail_id"], kwargs["mail_response_subject"]
+                    )
```
saruniitr, line 43: Instead of relying on presence/absence of values in arguments, consider sending the mail type to differentiate.
I think once you only refactor the sending of email this may not even be necessary except you may only need to know if we want to update status or not.

mail/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+                smtp_send(multipart_msg)
+                logger.info("Successfully sent email.")
+            except SMTPException as e:
+                logger.error(f"Failed to send email: {e}")
+                raise
+        else:
+            logger.info("Another send_email_task is currently in progress, will retry...")
+
+            retry_delay = RETRY_BACKOFF * (2**self.request.retries)
+            raise self.retry(countdown=retry_delay)
```
saruniitr, line 55: Do we really need to do this?
If the lock is not available does celery not retry automatically?

mail/tests/test_celery_task.py
```diff
@@ -1,36 +1,9 @@
...
-            "MIME-Version": "1.0",
-            "From": "test@example.com",  # /PS-IGNORE
-            "To": "notify@example.com",  # /PS-IGNORE
-            "Subject": "Licence rejected by HMRC",
-        }
-        self.assertDictEqual(dict(message), expected_headers)
-
-        text_payload = message.get_payload(0)
-        expected_body = "Mail (Id: 123) with subject CHIEF_SPIRE_licenceReply_202401180900_42557 has rejected licences"
-        self.assertEqual(text_payload.get_payload(), expected_body)
```
saruniitr, line 33: This need not be removed.

mail/tasks.py
seijihg: the circular import errors was the reason I put in tasks its not something easily can be put in the celery_tasks

saruniitr: I think it is not going to be an issue after these changes.

mail/models.py
seijihg: moved back!

mail/celery_tasks.py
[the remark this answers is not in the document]
seijihg: created `send_smtp_task(self, multipart_msg) `which should do the job.

mail/tasks.py
seijihg: from what I have read it does not retry automatically. We need to raise `self.retry()`

mail/tasks.py
seijihg: refactored

mail/libraries/routing_controller.py
seijihg: This part is actually very tricky to test mock due to Celery async nature. And because of the circular import of the I had to import send_smtp_task inside the method. It made impossible for me to mock it. 

the idea was to move/use update_mail inside `send` and when the task `state` was SUCCESS. Due to nature of it it will need more time to work on which I dont have/ more precisely I could not work out. None of the Vault values worked for the smtp email and password for me to do it naturally without mocking.

As for now it will work exactly the same how it was. It will be locked to one task at a time.

mail/celery_tasks.py
```diff
@@ -98,7 +103,7 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
         body = MIMEText(f"Mail (Id: {mail_id}) with subject {mail_response_subject} has rejected licences")
         multipart_msg.attach(body)
 
-        smtp_send(multipart_msg)
+        send_smtp_task(multipart_msg)
 
     except SMTPException:
```
currycoder, line 108: We now will not raise this exception as the sending will be an asynchronous task - I think this try block can be factored out

mail/celery_tasks.py
```diff
@@ -273,3 +278,36 @@ def manage_inbox():
...
+                logger.info("Successfully sent email.")
+                if mail and email_message_dto:
+                    update_mail(mail, email_message_dto)
+            except SMTPException as e:
+                logger.error(f"Failed to send email: {e}")
+                raise
+        else:
+            logger.info("Another send_smtp_task is currently in progress, will retry...")
+
+            retry_delay = RETRY_BACKOFF * (2**self.request.retries)
```
currycoder, line 312: `RETRY_BACKOFF` is a huge 180S by default... Would suggest that the value for this (given you are doing an exponential backoff) is 10 seconds.

Also you don't need to do your own exponential calculation, see the docs; https://docs.celeryq.dev/en/stable/userguide/tasks.html#Task.retry_backoff

In short, I think you can just set the `retry_backoff=<int>` up in the shared_task decorator above - that will handle exponential retries based on your starting number.

mail/celery_tasks.py
```diff
@@ -273,3 +278,36 @@ def manage_inbox():
...
+    timeout_at = time.monotonic() + LOCK_EXPIRE - 3
+    status = cache.add(lock_id, "locked", LOCK_EXPIRE)
+    try:
+        yield status
+    finally:
+        if time.monotonic() < timeout_at and status:
+            cache.delete(lock_id)
+
+
+@shared_task(bind=True, autoretry_for=(SMTPException,), max_retries=MAX_ATTEMPTS, retry_backoff=RETRY_BACKOFF)
```
currycoder, line 294: It would be good to have an exception raised if this task finally fails after all of the retries are exhausted.  I've done that using a base Task class, here's an example; https://github.com/uktrade/lite-hmrc/blob/e6a3be49ff019ef856c4e23651839861ddf9eea0/mail/celery_tasks.py#L114

mail/celery_tasks.py
```diff
@@ -273,3 +278,36 @@ def manage_inbox():
...
+def send_smtp_task(self, multipart_msg, mail=None, email_message_dto=None):
+    global_lock_id = "global_send_email_lock"
+
+    with memcache_lock(global_lock_id) as acquired:
+        if acquired:
+            logger.info("Global lock acquired, sending email")
+            try:
+                smtp_send(multipart_msg)
+                logger.info("Successfully sent email.")
```
saruniitr, line 303: Currently this is a generic task to send emails so to know which mail we are sending from logs, suggest adding `subject` to this log.

mail/celery_tasks.py
```diff
@@ -98,7 +103,7 @@ def notify_users_of_rejected_licences(mail_id, mail_response_subject):
         body = MIMEText(f"Mail (Id: {mail_id}) with subject {mail_response_subject} has rejected licences")
         multipart_msg.attach(body)
 
-        smtp_send(multipart_msg)
+        send_smtp_task(multipart_msg)
```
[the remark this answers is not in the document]
seijihg: I have used `apply_async` here so I can pass `kwargs` in

mail/celery_tasks.py
```diff
@@ -273,3 +278,36 @@ def manage_inbox():
             exc_info=True,
         )
         raise exc
+
+
+@contextmanager
+def memcache_lock(lock_id, oid=None):
```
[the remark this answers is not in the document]
seijihg: done changed :D

mail/libraries/routing_controller.py
```diff
@@ -168,10 +169,13 @@ def update_mail(mail: Mail, mail_dto: EmailMessageDto):
...
 
-def send(email_message_dto: EmailMessageDto):
+def send(email_message_dto: EmailMessageDto, mail=None):
+    # Needs to be imported here otherwise you will get a circular import errors.
+    from mail.celery_tasks import send_smtp_task
+
     logger.info("Preparing to send email")
     message = build_email_message(email_message_dto)
-    smtp_send(message)
+    send_smtp_task(message, mail, email_message_dto)
```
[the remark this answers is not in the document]
seijihg: now its apply_async so should be async

mail/celery_tasks.py
```diff
@@ -273,3 +278,36 @@ def manage_inbox():
...
+    timeout_at = time.monotonic() + LOCK_EXPIRE - 3
+    status = cache.add(lock_id, "locked", LOCK_EXPIRE)
```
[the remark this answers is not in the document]
seijihg: done from locked to lock_acquired

mail/celery_tasks.py
seijihg: remove all of them the new `base` should raise whatever is there on failures

mail/celery_tasks.py
seijihg: updated with base with `SendSmtpFailureTask`

mail/celery_tasks.py
```diff
@@ -3,6 +3,9 @@
 from email.mime.text import MIMEText
 from smtplib import SMTPException
 from typing import List, MutableMapping, Tuple
+import time
```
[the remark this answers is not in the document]
seijihg: done thanks!

mail/celery_tasks.py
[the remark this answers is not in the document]
seijihg: removed

mail/celery_tasks.py
seijihg: done added id as well

PR #232, opened 2024-02-08
[LTD-4684] Ensure non critical healthchecks do not result in a failed health check response

currycoder:
This change re-instates the separation of critical/non-critical health checks.  Non-critical health checks will still report an error through to sentry, but will not cause our pingdom checks to alert.  This will reduce alerting spam for the team.

### DIFF
--- a/healthcheck/checks.py
+++ b/healthcheck/checks.py
@@ -4,13 +4,12 @@ import poplib
 
 from django.conf import settings
 from django.utils import timezone
-
 from health_check.backends import BaseHealthCheckBackend
 from health_check.exceptions import HealthCheckException
+
 from mail.enums import ReceptionStatusEnum
 from mail.libraries.routing_controller import get_hmrc_to_dit_mailserver, get_spire_to_dit_mailserver
 from mail.models import LicencePayload, Mail
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


healthcheck/checks.py
```diff
@@ -28,35 +27,30 @@ def check_status(self):
...
 class LicencePayloadsHealthCheck(BaseHealthCheckBackend):
+    critical_service = False
+
     def check_status(self):
         dt = timezone.now() + datetime.timedelta(seconds=settings.LICENSE_POLL_INTERVAL)
         unprocessed_payloads = LicencePayload.objects.filter(is_processed=False, received_at__lte=dt)
 
         for unprocessed_payload in unprocessed_payloads:
             error_message = f"Payload object has been unprocessed for over {settings.LICENSE_POLL_INTERVAL} seconds: {unprocessed_payload}"
-            logger.error(error_message)
```
kevincarrogan, line 44: We do want these `logging.error` calls to happen, unless `self.add_error(HealthCheckException(error_message))` itself is manifesting in a Sentry log.

currycoder: I believe it is, but that's one of the things I'm aiming to check in putting it on an environment.  I think right now there's a lot of duplicate logs that happen all at once so this is an effort to sort that out

currycoder: Confirmed that these still show up in sentry despite the removed lines

seijihg (approved):

healthcheck/checks.py
seijihg: just out of curiosity can I get roughly idea how did you check it? just modified the code to output errors and check in sentry?

healthcheck/checks.py
```diff
@@ -28,35 +27,30 @@ def check_status(self):
...
             self.add_error(HealthCheckException(error_message))
 
-        if unprocessed_payloads.exists():
-            raise HealthCheckException("There are unprocessed licence payloads.")
-
 
 class PendingMailHealthCheck(BaseHealthCheckBackend):
+    critical_service = False
+
```
saruniitr, line 49: It can be argued that this is critical in the case if we have to resend any mails or based on mail from SPIRE team we may have to ask HMRC to look into a particular run number.
By marking it as otherwise the chances of ignoring are higher.

saruniitr (approved):

PR #233, opened 2024-02-13
LTD-4652: Manage concurrent SMTP connections using a lock

saruniitr:
## Change description

We use SMTP to send emails. As we process messages we also have a requirement to send emails from multiple tasks, because of this we may open multiple SMTP connections concurrently. This results in error if the number of concurrent connections exceed maximum allowed value.

To manage this reliably we are restricting access to this shared resource using a lock. This is achieved by deferring all email sending functionality to a dedicated task.
Before sending email it first tries to acquire a lock.
  - If there are no active connections then it acquires lock and sends email.
     In some cases we need to update state which is handled in subtasks linked to the parent task.
  - If there is active connection (lock acquisition fails) then it raises an exception to trigger an automatic retry.
     If all retries fail then manual intervention may be required (unlikely)

`Co-authored-by: Le H Ngo <hgle.ngo@gmail.com>`

[LTD-4652](https://uktrade.atlassian.net/browse/LTD-4652)

[LTD-4652]: https://uktrade.atlassian.net/browse/LTD-4652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mail/celery_tasks.py
+++ b/mail/celery_tasks.py
@@ -220,11 +220,11 @@ def finalise_sending_lite_licence_details(mail_id, message_dto, licence_payload_
     update_mail(mail, message_dto)
 
     licence_payloads = LicencePayload.objects.filter(id__in=licence_payload_ids, is_processed=False)
-    if licence_payloads:
-        licence_payloads.update(is_processed=True)
+    references = [item.reference for item in licence_payloads]
 
-        references = licence_payloads.values_list("reference", flat=True)
-        logger.info("Licence payloads of references %s marked as processed", references)
+    licence_payloads.update(is_processed=True)
+
+    logger.info("Licence payloads with references %s marked as processed", references)
 
 
 class SendLicenceDetailsBaseTask(Task):
--- a/mail/celery_tasks.py
+++ b/mail/celery_tasks.py
@@ -14,7 +14,7 @@ from django.utils import timezone
 from rest_framework.status import HTTP_207_MULTI_STATUS, HTTP_208_ALREADY_REPORTED
 
 from mail import requests as mail_requests
-from mail.enums import EmailType, ReceptionStatusEnum, SourceEnum
+from mail.enums import ReceptionStatusEnum, SourceEnum
 from mail.libraries.builders import build_email_message, build_licence_data_mail, build_licence_rejected_email_message
 from mail.libraries.data_processors import build_request_mail_message_dto
 from mail.libraries.email_message_dto import EmailMessageDto
@@ -103,27 +103,6 @@ def cache_lock(lock_id):
 
 
 class SendEmailBaseTask(Task):
-    def on_success(self, retval, task_id, args, kwargs):
-        logger.info("Sending email task successful")
-
-        email_type, mail_id, message_dto = args
-        if email_type == EmailType.LITE_LICENCE_DETAILS or email_type == EmailType.SPIRE_LICENCE_DETAILS:
-            message_dto = EmailMessageDto(*message_dto)
-            mail = Mail.objects.get(id=mail_id)
-            update_mail(mail, message_dto)
-            logger.info("Mail %s status updated to %s", mail.edi_filename, mail.status)
-
-        licence_payload_ids = kwargs.get("licence_payload_ids", [])
-        if email_type == EmailType.LITE_LICENCE_DETAILS and licence_payload_ids:
-            # Mark payloads as processed
-            licence_payloads = LicencePayload.objects.filter(id__in=licence_payload_ids)
-            licence_payloads.update(is_processed=True)
-
-            references = licence_payloads.values_list("reference", flat=True)
-            logger.info("Licence payloads of references %s marked as processed", references)
-
-        return super().on_success(retval, task_id, args, kwargs)
-
     def on_failure(self, exc, task_id, args, kwargs, einfo):
         message = """
         Maximum attempts for send_email_task exceeded - the task has failed and needs manual inspection.
@@ -142,9 +121,9 @@ class SendEmailBaseTask(Task):
     retry_backoff=RETRY_BACKOFF,
     base=SendEmailBaseTask,
 )
-def send_email_task(email_type, mail_id, message_dto, licence_payload_ids=None):
+def send_email_task(message):
     """
-    Main purpose of this task is to send email based on the type provided.
+    Main purpose of this task is to send email.
 
     We use SMTP to send emails. As we process messages we have a requirement to
     send emails from multiple places, because of this we may open multiple SMTP connections.
@@ -154,7 +133,7 @@ def send_email_task(email_type, mail_id, message_dto, licence_payload_ids=None):
     This is achieved by deferring all email sending functionality to this task.
     Before sending email it first tries to acquire a lock.
       - If there are no active connections then it acquires lock and sends email.
-        In some cases we need to update state which is handled in on_success() callback.
+        In some cases we need to update state which is handled in subtask linked to this task.
       - If there is active connection (lock acquisition fails) then it raises an exception
         which triggers a retry.
         If all retries fail then manual intervention may be required (unlikely)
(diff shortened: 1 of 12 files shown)


mail/libraries/routing_controller.py
```diff
@@ -175,6 +175,8 @@ def send(email_message_dto: EmailMessageDto):
 
 
 def _collect_and_send(mail: Mail):
+    from mail.celery_tasks import send_email_task
+
```
saruniitr, line 179: unavoidable due to circular dependencies

mail/tests/test_send_lite_licence_updates_task.py
```diff
@@ -1,49 +1,84 @@
...
-            (ReceptionStatusEnum.REPLY_PENDING, 1, 1, 0),
-            (ReceptionStatusEnum.REPLY_RECEIVED, 1, 1, 0),
-            (ReceptionStatusEnum.REPLY_SENT, 2, 1, 1),
+            ([True],),
+            ([False, True],),
+        ]
+    )
+    @mock.patch("mail.celery_tasks.cache")
+    @mock.patch("mail.celery_tasks.smtp_send")
+    def test_send_licence_details_success(self, lock_acquired, mock_smtp_send, mock_cache):
```
saruniitr, line 27: Ideally this test should be in `test_celery_tasks.py` but existing tests for this task are in a separate file so added it here.

saruniitr:
> I am not sure since ur are not injecting straight into `send` its still being used by `parse_and_reply_emails` and there is management command which is using but I guess this not matter as much, would not go directly into the send better as I did?

You mean call `send()` directly instead of below?
```
            if email_type == EmailType.LITE_LICENCE_DETAILS or email_type == EmailType.SPIRE_LICENCE_DETAILS:
                message = build_email_message(message_dto)
                smtp_send(message)

            elif email_type == EmailType.LICENCE_REJECTED:
                message = build_licence_rejected_email_message(message_dto)
                smtp_send(message)
```

Yes, we can call `send()` directly in the first `if` condition but to keep it consistent with the second `if` condition I am calling individual functions.

You can ignore `parse_and_reply_emails()` which part of `mock_hmrc` as it is incomplete. I think we can remove that completely.

Yes, `send()` is also called in a management command `resend_email()`, I considered changing that to a task as well but that is not required as we only use it manually in special circumstances.

mail/celery_tasks.py
```diff
@@ -75,38 +80,132 @@ def _log_error(message, lite_usage_data_id):
...
+            logger.exception("Another SMTP connection is active, will be retried after backing off")
+            raise SMTPConnectionBusy()
+
+        logger.info("Lock acquired, proceeding to send email")
+
+        # de-serialize back to original type
+        message_dto = EmailMessageDto(*message_dto)
+
+        message = None
+        if email_type == EmailType.LITE_LICENCE_DETAILS or email_type == EmailType.SPIRE_LICENCE_DETAILS:
```
saruniitr, line 176: this type is introduced to identify which task is sending this email.
Each task has a different set of arguments and follow-up actions after sending an email.
It may be possible to identify based on other arguments but made it explicit.

Eg the task that send licences rejected notification email only needs to send `message_dto`. It can send `mail_id` as `None` and we can use that to identify but that will introduce additional if conditions.

The other three instances where this is called is when SPIRE sends new licence details, reply back and usage data back - in all these cases we need `mail_id, dto` and update them as follow up tasks. 

The only task that provides `licence_payload_ids` is the one that sends LITE licence details hence it is assigned a default value so other 3 tasks can ignore this.

saruniitr: We can try to pass the final message to be sent as argument so this task would simply call `smtp_send(message)` but we still need `Mail, Dto` arguments to perform follow-up actions.

mail/tests/test_resend_email.py
```diff
@@ -62,19 +69,26 @@ def test_resend_licence_data_mail_to_hmrc(
         self.assertEqual(mail.id, pending_mail.id)
         self.assertEqual(mail.status, ReceptionStatusEnum.REPLY_PENDING)
         self.assertEqual(mail.extract_type, ExtractTypeEnum.LICENCE_DATA)
-        self.assertEqual(send_mail.call_count, 2)
+        self.assertEqual(mock_send.call_count, 1)
 
+    @override_settings(SEND_REJECTED_EMAIL=False)
```
saruniitr, line 74: this just interferes with the call count value because the reply message content has rejections so disabled this as it is not related to this particular test.

hnryjmes:
@saruniitr has this been tested on a dev / demo environment?

mail/celery_tasks.py
```diff
@@ -75,38 +80,132 @@ def _log_error(message, lite_usage_data_id):
...
+        licence_payload_ids = kwargs.get("licence_payload_ids", [])
+        if email_type == EmailType.LITE_LICENCE_DETAILS and licence_payload_ids:
+            # Mark payloads as processed
+            licence_payloads = LicencePayload.objects.filter(id__in=licence_payload_ids)
+            licence_payloads.update(is_processed=True)
+
+            references = licence_payloads.values_list("reference", flat=True)
+            logger.info("Licence payloads of references %s marked as processed", references)
+
+        return super().on_success(retval, task_id, args, kwargs)
```
currycoder, line 125: We can do callbacks with celery, which might make this a bit cleaner.  https://docs.celeryq.dev/en/stable/userguide/calling.html#linking-callbacks-errbacks
Gurdeep implemented something similar here; https://github.com/uktrade/lite-api/pull/1811/files#diff-ba75c0a31a6b6ed70ec3adffebfd6843645ab9dde64b35fe1c49092919e2857dR22

currycoder: Benefit of this is that we can have the "success operations" defined by the caller to `send_email_task` rather than baked in to here with a switch statement

mail/celery_tasks.py
```diff
@@ -75,38 +80,132 @@ def _log_error(message, lite_usage_data_id):
...
+            message = build_licence_rejected_email_message(message_dto)
+
+        try:
+            if message:
+                smtp_send(message)
+        except SMTPException:
+            logger.exception("An unexpected error occurred when sending email -> %s")
+            raise
+
+        logger.info("Email sent successfully")
```
hnryjmes, line 188: is `message` is None it looks like you will still get to the `logger.info` statement 

wouldn't you want the success message to be inside the try block and the failure message to be the fall through option?

saruniitr: I will move this to within the if condition.

mail/celery_tasks.py
currycoder: I realise you're heading this comment off, but this pattern where we need the guts of our email sending to do conditional things to build/finalise an email is a bit of a shame.  Did you try doing the custom serialization?  What made it difficult?

hnryjmes (approved):

mail/celery_tasks.py
saruniitr: I have not tried that as using the type that can be avoided.
We need custom serialization for two models but with linking the tasks it may also be not necessary.

saruniitr: @currycoder updated to use task specific serializer.

mail/celery_tasks.py
```diff
@@ -75,38 +80,106 @@ def _log_error(message, lite_usage_data_id):
...
 
 
-# Notify Users of Rejected Mail
+class SMTPConnectionBusy(SMTPException):
+    pass
+
+
+@contextmanager
+def cache_lock(lock_id):
+    timeout_at = time.monotonic() + LOCK_EXPIRE - 3
```
currycoder, line 94: what's the `- 3` bit doing?

currycoder (approved):
Approved, pending code coverage - just one question for my understanding

mail/celery_tasks.py
saruniitr: I am also not sure, this is taken from the example in Celery documentation.

PR #234, opened 2024-02-14
LTD-4630: Remove references to process_tasks management command

saruniitr:
## Change description

This is a management command in background_task which is used to start the tasks. We have completely removed background_tasks and switched to celery so this are no longer required.


depsiatwal (approved):
lgtm

currycoder (approved):

PR #238, opened 2024-03-04
LTD-4775: Add capability to dump anonymised db sql file

saruniitr:
## Change description

Objective is to replace mechanism of generating anonymised DB dump for lite-hmrc. Currently this is done by an airflow pipeline which is not easy to update/maintain.
This PR follows the same pattern as in `lite-api` and adds a management command `dump_and_anonymise` that writes anonymised db sql file to a S3 bucket. The command itself is in a submodule.

The fields to anonymise are specified in a yaml file `anonymise_mode_config.yaml` which is in this repo. The methods that specify the anonymising behaviour for each of those fields in the yaml file are in the submodule. Currently it reproduces the like-for-like behaviour of existing db anonymisation process but follow up work is going to extend such that we get a more realistic anonymised data and replacing only the bits that contains sensitive data.

Additional environment variables related to AWS S3 are added as currently no buckets are bound to this service. We need to ensure these are configured when this is ready to deployed.

Unit tests ensure data is anonymised as expected by checking absence of data in the anonymised sql file.


conf/settings.py
```diff
@@ -291,3 +294,30 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+            DB_ANONYMISER_AWS_ACCESS_KEY_ID = aws_credentials["aws_access_key_id"]
+            DB_ANONYMISER_AWS_SECRET_ACCESS_KEY = aws_credentials["aws_secret_access_key"]
+            DB_ANONYMISER_AWS_REGION = aws_credentials["aws_region"]
+            DB_ANONYMISER_AWS_STORAGE_BUCKET_NAME = aws_credentials["bucket_name"]
+else:
+    DB_ANONYMISER_AWS_ENDPOINT_URL = AWS_ENDPOINT_URL
+    DB_ANONYMISER_AWS_ACCESS_KEY_ID = env("DB_ANONYMISER_AWS_ACCESS_KEY_ID", default=None)
+    DB_ANONYMISER_AWS_SECRET_ACCESS_KEY = env("DB_ANONYMISER_AWS_SECRET_ACCESS_KEY", default=None)
+    DB_ANONYMISER_AWS_REGION = env("DB_ANONYMISER_AWS_REGION", default=None)
+    DB_ANONYMISER_AWS_STORAGE_BUCKET_NAME = env("DB_ANONYMISER_AWS_STORAGE_BUCKET_NAME", default=None)
```
saruniitr, line 320: this can probably be removed as no S3 buckets were previously associated with lite-hmrc.

bin/dump_and_anonymise.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
saruniitr, line 1: this is required for `pg_dump` to correctly work on PaaS.
Taken from a similar change in `lite-api` https://github.com/uktrade/lite-api/pull/1834/files

saruniitr:
Verified on dev environment and working as expected.

conf/settings.py
```diff
@@ -291,3 +294,30 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+    for bucket_details in VCAP_SERVICES["aws-s3-bucket"]:
+        if S3_BUCKET_TAG_ANONYMISER_DESTINATION in bucket_details["tags"]:
+            aws_credentials = bucket_details["credentials"]
+            DB_ANONYMISER_AWS_ENDPOINT_URL = None
+            DB_ANONYMISER_AWS_ACCESS_KEY_ID = aws_credentials["aws_access_key_id"]
+            DB_ANONYMISER_AWS_SECRET_ACCESS_KEY = aws_credentials["aws_secret_access_key"]
+            DB_ANONYMISER_AWS_REGION = aws_credentials["aws_region"]
+            DB_ANONYMISER_AWS_STORAGE_BUCKET_NAME = aws_credentials["bucket_name"]
+else:
+    DB_ANONYMISER_AWS_ENDPOINT_URL = AWS_ENDPOINT_URL
```
depsiatwal, line 316: just thinking that can this be removed ?
Assume we don't want to bind through vault

anonymised_db_dumps/tests/test_anonymised_dumps.py
```diff
@@ -0,0 +1,61 @@
...
+        cls.dump_location = f"/tmp/{settings.DB_ANONYMISER_DUMP_FILE_NAME}"
+        try:
+            os.remove(cls.dump_location)
+        except FileNotFoundError:
+            pass
+
+        call_command("dump_and_anonymise", keep_local_dumpfile=True, skip_s3_upload=True)
+
+        with open(cls.dump_location, "r") as f:
+            cls.anonymised_sql = f.read()
```
currycoder, line 26: I have an improved pattern for these testcases in lite-api.  The original pattern just asserted the presence/absence of strings from the dumped SQL string.  The new pattern actually loads the dump back in to the test database and asserts the presence/absence of strings on the model records themselves.  See; https://github.com/uktrade/lite-api/pull/1850/commits/3dfbfb8e8fd4f0010ad3e0f2483bcc680aa79b40

conf/settings.py
```diff
@@ -291,3 +294,30 @@ def _build_redis_url(base_url, db_number, **query_args):
...
 CELERY_TASK_SEND_SENT_EVENT = env.bool("CELERY_TASK_SEND_SENT_EVENT", True)
+
+
+S3_BUCKET_TAG_ANONYMISER_DESTINATION = "anonymiser"
+
+AWS_ENDPOINT_URL = env("AWS_ENDPOINT_URL", default=None)
+
+if VCAP_SERVICES:
+    if "aws-s3-bucket" not in VCAP_SERVICES:
+        raise Exception("S3 Bucket not bound to environment")
```
currycoder, line 305: Would suggest dropping this for lite-hmrc so that we can have environments working that do not have the S3 bucket.  IMO the bucket is only really needed for testing anonymisation.

currycoder: Another reason that I suggest this is that it will make the deployment to production easier - so that we do not strictly have to create/bind the new S3 bucket before deploying.

conf/settings.py
currycoder: This is more for local development

conf/settings.py
currycoder: I don't follow - what can be removed?

conf/anonymise_model_config.yaml
```diff
@@ -0,0 +1,12 @@
...
+    - django_db_anonymiser.db_anonymiser
+    - mail
+
+strategy:
+  mail:
+    raw_data: anonymisers.raw_data
+    edi_data: anonymisers.edi_data
+    sent_data: anonymisers.sent_data
+  mail_licencepayload:
+    data: anonymisers.payload_data
```
currycoder, line 12: I noticed that in the old anonymiser code we also anonymised `rpa_trader_id`; https://github.com/uktrade/lite-airflow-dags/blob/master/dags/lite_hmrc_model_config.yaml#L7

Do we need to reflect that here too?

saruniitr: That model is not used anywhere in the code. I have also checked that there are no objects created for that model hence removed it.

currycoder (approved):

PR #239, opened 2024-03-07
LTD-4793: Fix failure when sending licence rejected notifications email

saruniitr:
## Change description

If the licence details sent to HMRC doesn't conform to their spec then they reject and inform us in licence reply message. `lite-hmrc` detects and sends email notifications to users (if `SEND_REJECTED_EMAIL=True`) for visibility.

Previously the task to send this email is scheduled from `Mail.save()` method. This works but the issue is when processing licence reply we update various fields which call `save()` again creating multiple tasks (4) to send the same email.

As all of these try to acquire SMTP connection to send email it was exceeding max concurrent connections resulting in failure. This was fixed by protecting this shared resource using a lock so only one connection is active at a time. It solved the concurrent connections issue but didn't address the underlying issue of creating multiple tasks for the same email.

We also used a custom serializer (`pickle`) to serialize the argument (message of type `MIMEMultipart()`) sent to this task as it is not JSON serializable and it was specified at the time of scheduling the task.

All the tasks were serialized as expected but they were failing on retry (if they don't get a lock they retry). We raise an exception to trigger `autoretry_for` for this purpose. The issue when they retry was, task restarts and it was using a default serializer which is `json`. For this reason retry was being rejected.
Another main issue is that the processing of licence reply was also not complete (this is one of the task waiting for lock) so it is stuck in `reply_received` status essentially blocking the queue.

Below are the changes made in this PR to fix these issues,
 - Move scheduling of licence rejected notification away from `save()` and do this after completely processing licence reply. This won't leave things in inconsistent state.
 - Specify `pickle` serializer using the task decorator. With this even if it restarts during a retry it is guaranteed to use the correct serializer always.
 - Add test for this specific case where licences are rejected and ensure correct number of tasks are scheduled.

### Testing on environment
All of following cases are verified on UAT with these changes and working as expected.
- Sending of licence data from SPIRE
- Processing reply message for the above with rejected licences
- Sending of LITE issued licence details
- Processing of usage data


mail/libraries/routing_controller.py
```diff
@@ -216,3 +217,13 @@ def get_email_message_dtos(server: MailServer, number: Optional[int] = 3) -> Lis
     # emails = read_last_three_emails(pop3_connection)
     server.quit_pop3_connection()
     return emails
+
+
+def check_and_notify_rejected_licences(mail):
+    from mail.celery_tasks import notify_users_of_rejected_licences
```
saruniitr, line 223: unavoidable due to circular dependencies.

mail/celery_tasks.py
```diff
@@ -120,6 +120,7 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
     max_retries=MAX_ATTEMPTS,
     retry_backoff=RETRY_BACKOFF,
     base=SendEmailBaseTask,
+    serializer="pickle",
```
saruniitr, line 123: this is the main change

mail/celery_tasks.py
```diff
@@ -146,19 +147,18 @@ def send_email_task(message):
             logger.exception("Another SMTP connection is active, will be retried after backing off")
             raise SMTPConnectionBusy()
 
-        logger.info("Lock acquired, proceeding to send email")
+        logger.info("Lock acquired, proceeding to send email from %s to %s", message["From"], message["To"])
 
         try:
             smtp_send(message)
         except SMTPException:
             logger.exception("An unexpected error occurred when sending email -> %s")
```
depsiatwal, line 155: i'm wondering do we need to log at generic exception level to say it wasn't successful ?

mail/celery_tasks.py
```diff
@@ -146,19 +147,18 @@ def send_email_task(message):
...
             raise
 
-        logger.info("Email sent successfully")
+        logger.info("Email sent successfully to %s", message["To"])
```
depsiatwal, line 158: does message["To"] contain any PII info if so we may want to avoid for this reason.

mail/celery_tasks.py
saruniitr: Yes, it is the base exception for various SMTP errors so logs helps to understand why it failed.
This is the same pattern we are following in all tasks.

depsiatwal: Yeah  I mean do we need another catch so we log if it wasn't SMTPException error

mail/celery_tasks.py
saruniitr: It is a mail id and it is going to ELK. Even if we remove it here the same details are logged in other places.
The main reason I added here is because this task is called by multiple users and there is nothing differentiating to know which email is being sent to whom other than From/To.

mail/libraries/routing_controller.py
```diff
@@ -216,3 +217,13 @@ def get_email_message_dtos(server: MailServer, number: Optional[int] = 3) -> Lis
...
+
+    if not settings.SEND_REJECTED_EMAIL:
```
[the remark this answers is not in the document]
depsiatwal: fair enough

mail/celery_tasks.py
saruniitr: no, I think this should be enough

depsiatwal: so if the task failed for a reason which wasn't SMTPException how would we know by looking at the logs ?

saruniitr: I think for me the fact that it logs something to ELK is more important here especially as there are multiple users for this task.

mail/libraries/routing_controller.py
hnryjmes: I thought for a while about whether the `notify_users_of_rejected_licences` function could be moved somewhere or even brought entirely into the `routing_controller` file but ultimately I think what you've got currently is fine

hnryjmes (approved):

depsiatwal (approved):
LGTM had a comment around logging but no big deal if you think its fine.

currycoder (approved):
Great PR description 🙌

mail/celery_tasks.py
saruniitr: I am going to leave this as is as try catch is only for the `smtp_send()` and `SMTPException` covers all cases.

PR #240, opened 2024-03-15
LTD-4776: Review and update anonymisation of model fields

saruniitr:
## Change description

The licence details that we send to HMRC contain PI data and these are present in various fields as the mail object gets updated during its processing.
These fields are,
- `Mail` model: `edi_data`, `sent_data`, `raw_data`
- `LicencePayload` model: `data`

Currently we are replacing all of their contents with a static text. This is not very helpful if we need to check any of these details for debug. Since PI data is only present in specific lines of message it is better to anonymise only those lines.

This is now possible because we have anonymiser functions in python itself and can be configured for each field.

In this PR these functions are updated to parse through `edi_data` and clear PI data keeping the rest of it as is.
`sent_data` and `raw_data` can be completely masked out as it is the same as `edi_data`.

LicencePayload anonymiser also update to include some details about the payload.

Add tests to verify relevant lines are anonymised as expected.

#### Testing on environment
verified on devdata and working as expected


anonymised_db_dumps/tests/test_anonymised_dumps.py
```diff
@@ -71,30 +74,180 @@ def tearDownClass(cls):
...
+                "foreignTrader",
+                "%s\\foreignTrader\\End-user name\\address line1\\address line2\\address line3\\address line4\\address line5\\postcode\\AU",
+            ),
+            (
+                "line",
+                "%s\\line\\1\\\\\\\\\\PRODUCT NAME\\Q\\\\030\\\\1\\\\\\\\\\\\",
+            ),
+        ]
+    )
+    def test_mail_with_siels_unit_NAR_anonymised(self, chief_line_type, expected_line_format):
```
saruniitr, line 138: There is duplication in the test code which is unavoidable.
The format of `trader` and `foreignTrader` lines are same for all licence types but the product line format is different.
For SIELs we have quantity and units which results in multiple combinations.
For Open licences there is no quantity.

So to handle both in the same test results in multiple if conditions etc, to avoid that I have separated out tests for each licence type resulting in duplication.

saruniitr:
Verified on `devdata` and working as expected.

Tllew (approved):

hnryjmes (approved):

PR #242, opened 2024-03-26
LTD-4776: Skip empty or invalid lines when anonymising edi_data

saruniitr:
## Change description

Mail `edi_data` is anonymised based on line type which is derived from the line content however if line is empty or invalid then we won't get expected number of tokens and anonymisation fails. In such case skip those lines.

To keep things simple we are not adding any static text when we skip lines because this can be at the end after `fileTrailer` so adding any content after that does not make sense.


mail/anonymisers.py
```diff
@@ -63,12 +63,15 @@ def sanitize_product_line(line):
...
-    if "fileHeader" not in lines:
+    if "fileHeader" not in lines and "fileTrailer" not in lines:
         return f"{today()}: invalid edi data"
 
     output_lines = []
     for line in lines.split("\n"):
         tokens = line.split("\\")
+        # skip empty or invalid lines
+        if len(tokens) < 2:
+            continue
```
currycoder, line 74: Does anything in the test data trigger this particular statement?

saruniitr: yes the newline in the above test file triggers this. Verified with a breakpoint.
To make it more clear should I add a dummy line in between?

saruniitr: I have added a separate test for this with invalid content.

hnryjmes (approved):

currycoder (approved):

PR #259, opened 2024-06-24
Use custom code coverage checker

kevincarrogan:
This removes us using the external codecov tool, which for some reason was no longer working as expected, with the same toolset that we use on our other repos.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,10 +48,16 @@ jobs:
 
       - run:
           name: Run tests
-          command: pipenv run pytest --cov=. --cov-report xml --cov-config=.coveragerc
+          command: pipenv run pytest --cov=. --cov-config=.coveragerc
 
-      - store_test_results:
-          path: test-results
+      - run:
+          name: Rename coverage file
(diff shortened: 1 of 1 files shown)


markj0hnst0n (approved):

.circleci/config.yml
```diff
@@ -164,6 +151,39 @@ jobs:
           # the cache key changes whenever any of the branch name, Pipfile.lock checksum, or .pre-commit-config.yaml checksum change
           key: dependencies-{{ .Branch }}-{{ checksum "Pipfile.lock" }}-{{ checksum ".pre-commit-config.yaml" }}
 
+  check_coverage:
+    working_directory: ~/lite-hmrc
+    docker:
+      - image: cimg/python:3.8.17
```
hnryjmes, line 157: is there a reason this has to be `3.8.17`? I notice the other python images in the file and the python version in the Pipfile are at `3.8.18`

kevincarrogan: Copy and paste issue. Will update.

hnryjmes (approved):

PR #262, opened 2024-06-25
 LTD-5138-dbt-application-settings 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5138


conf/settings.py
```diff
@@ -319,12 +276,113 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+    if REDIS_BASE_URL:
+        # Give celery tasks their own redis DB - future uses of redis should use a different DB
+        REDIS_CELERY_DB = env("REDIS_CELERY_DB", default=0)
+        is_redis_ssl = REDIS_BASE_URL.startswith("rediss://")
+        url_args = {"ssl_cert_reqs": "CERT_REQUIRED"} if is_redis_ssl else {}
+
+        CELERY_BROKER_URL = _build_redis_url(REDIS_BASE_URL, REDIS_CELERY_DB, **url_args)
+        CELERY_RESULT_BACKEND = CELERY_BROKER_URL
+
+        CACHES = {
```
Tllew, line 375: feels like there's a lot of duplication in this, could we set defaults before the if elif?

depsiatwal: Yeah that was kind of deliberate the idea is we will just deleted everything from PAAS Platform eventually making this a lot tidier.

markj0hnst0n (approved):

PR #265, opened 2024-06-28
LTD-3688-add/amend-units-of-measurement

markj0hnst0n:
### Aim

I've made notes of the prod data impact with comments on the tickets relating to these issues.

Make the following amendments:

- Tonnes / TON - Add and use code 25
[LTD-3690](https://uktrade.atlassian.net/browse/LTD-3690)

- Milligrams / MIM - use code 111 and change to “MGM”
- Micrograms / MCM - use code 116 and change to “MCG”
- Microlitres / MCR - use code 110 and change to “MCL”
[LTD-3688](https://uktrade.atlassian.net/browse/LTD-3688)

- Millilitres / MIR - change to “MLT”
[LTD-3691](https://uktrade.atlassian.net/browse/LTD-3691)

Corresponding LITE API PR [here](https://github.com/uktrade/lite-api/pull/2066)

[LTD-3688]: https://uktrade.atlassian.net/browse/LTD-3688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-3690]: https://uktrade.atlassian.net/browse/LTD-3690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-3691]: https://uktrade.atlassian.net/browse/LTD-3691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -205,6 +205,7 @@ class UnitMapping(enum.Enum):
         # Used by the API serializer for validation.
         return list(cls.__members__.keys())
 
+
 class LegacyUnitCodeMapping(enum.Enum):
     # These units below are the old values that are not used going forward
     # They are kept here for backwards compatibility
@@ -212,6 +213,7 @@ class LegacyUnitCodeMapping(enum.Enum):
     MCM = 116  # microgram
     MIR = 74  # millilitre
     MCR = 110  # microlitre
+
 
 class MailReadStatuses(TextChoices):
     READ = "READ"
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -219,6 +219,8 @@ class LegacyUnitMapping(enum.Enum):
         # Used by the API serializer for validation.
         return list(cls.__members__.keys())
 
+    def get_class_attributes(cls):
+        return [attr for attr in dir(cls)]
 
 class MailReadStatuses(TextChoices):
     READ = "READ"
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -214,6 +214,11 @@ class LegacyUnitCodeMapping(enum.Enum):
     MIR = 74  # millilitre
     MCR = 110  # microlitre
 
+    @classmethod
+    def serializer_choices(cls):
+        # Used by the API serializer for validation.
+        return list(cls.__members__.keys())
+
 
 class MailReadStatuses(TextChoices):
     READ = "READ"
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -206,7 +206,7 @@ class UnitMapping(enum.Enum):
         return list(cls.__members__.keys())
 
 
-class LegacyUnitCodeMapping(enum.Enum):
+class LegacyUnitMapping(enum.Enum):
     # These units below are the old values that are not used going forward
     # They are kept here for backwards compatibility
     MIM = 111  # milligram
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -200,15 +200,8 @@ class UnitMapping(enum.Enum):
     MCG = 116  # microgram
     MCL = 110  # microlitre
 
-    @classmethod
-    def serializer_choices(cls):
-        # Used by the API serializer for validation.
-        return list(cls.__members__.keys())
-
-
-class LegacyUnitMapping(enum.Enum):
-    # These units below are the old values that are not used going forward
-    # They are kept here for backwards compatibility
+    # These units below are the old values or LegacyUnitMapping that are
+    # not used going forward. They are kept here for backwards compatibility
     MIM = 111  # milligram
     MCM = 116  # microgram
     MIR = 74  # millilitre
@@ -218,9 +211,6 @@ class LegacyUnitMapping(enum.Enum):
     def serializer_choices(cls):
         # Used by the API serializer for validation.
         return list(cls.__members__.keys())
-
-    def get_class_attributes(cls):
-        return [attr for attr in dir(cls)]
 
 
 class MailReadStatuses(TextChoices):
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -222,6 +222,7 @@ class LegacyUnitMapping(enum.Enum):
     def get_class_attributes(cls):
         return [attr for attr in dir(cls)]
 
+
 class MailReadStatuses(TextChoices):
     READ = "READ"
     UNREAD = "UNREAD"
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -193,9 +193,12 @@ class UnitMapping(enum.Enum):
     MTR = 57  # meters
     LTR = 94  # litre
     MTQ = 2  # meters_cubed
-    MIR = 74  # millilitre
+    MLT = 74  # millilitre
     ITG = 30  # intangible
-    MIM = 111  # milligram
+    MGM = 111  # milligram
+    TON = 25  # tonne
+    MCG = 116  # microgram
+    MCL = 110  # microlitre
 
     @classmethod
     def serializer_choices(cls):
--- a/mail/enums.py
+++ b/mail/enums.py
@@ -205,6 +205,13 @@ class UnitMapping(enum.Enum):
         # Used by the API serializer for validation.
         return list(cls.__members__.keys())
 
+class LegacyUnitCodeMapping(enum.Enum):
+    # These units below are the old values that are not used going forward
+    # They are kept here for backwards compatibility
+    MIM = 111  # milligram
+    MCM = 116  # microgram
+    MIR = 74  # millilitre
+    MCR = 110  # microlitre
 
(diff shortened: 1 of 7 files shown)


hnryjmes (approved):

seijihg (approved):

mail/enums.py
```diff
@@ -193,9 +193,12 @@ class UnitMapping(enum.Enum):
...
     LTR = 94  # litre
     MTQ = 2  # meters_cubed
-    MIR = 74  # millilitre
+    MLT = 74  # millilitre
     ITG = 30  # intangible
-    MIM = 111  # milligram
+    MGM = 111  # milligram
+    TON = 25  # tonne
+    MCG = 116  # microgram
+    MCL = 110  # microlitre
```
currycoder, line 201: I think we possibly have a deploy sequencing/backward compatibility issue here.

Can you explain how this `UnitMapping` enum will be used?

saruniitr: Yes, this looks like a simple change but can have greater impact if not sequenced correctly.
The unit letters are for our purposes but when we send licence details to HMRC we send the corresponding code in the message.

Once this is merged and if we issue any licence that uses eg `MIR` units then it won't find mapping and can fail here. I think before we make this change we need to see if any products use the units that are changing to new ones. Looking at the ticket it looks like we are correcting this to use correct unit codes so hopefully there won't be any products.

I would also suggest changing only one unit at a time to reduce the impact and isolate any issues.
Best to update `lite-hmrc` first in a backward compatible way as the codes are only exposed here, then API and then finally remove any temporary changes in `lite-hmrc`

markj0hnst0n: My understanding is that an email to HMRC containing information about a licence  is built via a call from the API or via an email from SPIRE and this email will contain the unit of measurement for an item that the licence has been issue with regard to.  The unit of measurement relates to a code HMRC will assign to it and we've previously not have the correct codes for this on the API or HMRC apps.  The UnitMapping is used to assign a constant to the unit which contains the numerical code and alphabetical code and effectively in the email is is rendered  for HMRC purpose as the numerical code.

As such the API should be deployed as close to simultaneous as possible to negate any deploy sequencing issues.

Was thinking about re-adding the old codes to the system and mapping them to the correct numerical code to negate backwards compatibility issues?

hnryjmes: I think if these old incompatible units were being used regularly, licence data would be failing somewhere at the HMRC end and we would hear about it. This happened before, which is what triggered the last set of these changes around when we wrote the ticket, in May 2023. But I don't think they are being used regularly. However that only tells us about licences that have already been issued, not cases that are under review now.

Arun has already made a similar point, and I agree:

>  I think before we make this change we need to see if any products use the units that are changing to new ones. Looking at the ticket it looks like we are correcting this to use correct unit codes so hopefully there won't be any products.

Regarding backwards compatibility and deployments - we sometimes aren't in a position to ensure simultaneous releases of things like lite-api and lite-hmrc. So it's better where possible to code things in a way that removes the necessity for such a simultaneous release. We could also explore using feature switches if that would make things easier (then a change could be effected by just updating Vault and redeploying).

hnryjmes (changes requested):
I am generally happy with the code changes but before we continue to merge this in we should have:
- a good understanding of which if any PROD cases would be affected by this change (i.e. which cases in prod have goods making use of the units of measurement that will be removed)
- a release plan (can be informal) taking into account backwards compatibility if we are not able to release lite-api and lite-hmrc simultaneously

mail/enums.py
currycoder: I think my proposal would be;
- Make a change on lite-hmrc which supports both the new units AND the old units.  Old units should be commented as legacy in a very obvious way.  Push it all the way through to production.
- Make a change on lite-api which ensures it only sends the new units.  Push that all the way through to production.
- Optionally; make a change on lite-hmrc to drop any mention to the old units.  I think before we do that I'd like to try answer whether there is any scope to replay old messages with the old units or anything like that..

What do we think?  Would this cover the various scenarios we have discussed on this thread?

saruniitr: Yes exactly this. First point is the crucial one.
Please also see if any products use these units to get an idea of the scale of this.

mail/tests/test_licence_to_edifact.py
```diff
@@ -99,6 +105,7 @@ def test_update_edifact_file(self):
 
         trader = licences[0].data["organisation"]
         now = timezone.now()
+        self.maxDiff = None
```
hnryjmes, line 108: is this left over from your own testing?

mail/serializers.py
```diff
@@ -146,7 +146,10 @@ class GoodSerializer(serializers.Serializer):
     name = serializers.CharField()
     description = serializers.CharField(max_length=2000, allow_blank=True)
     quantity = serializers.DecimalField(decimal_places=3, max_digits=13)
-    unit = serializers.ChoiceField(choices=enums.UnitMapping.serializer_choices())
+    combined_unit_mapping_choices = (
+        enums.UnitMapping.serializer_choices() + enums.LegacyUnitMapping.serializer_choices()
```
hnryjmes, line 150: question for my own info - presumably here, if a key existed in both lists, then adding the lists here would cause an error? or are duplicate choices allowed? I can see that there aren't duplicate choices now with the units the way they are but it seems possible

hnryjmes (approved):
looks better now with the backwards compatibility. given that we've agreed a release plan now, happy to approve

mail/enums.py
```diff
@@ -193,9 +193,26 @@ class UnitMapping(enum.Enum):
...
+        return list(cls.__members__.keys())
+
+
+class LegacyUnitMapping(enum.Enum):
+    # These units below are the old values that are not used going forward
+    # They are kept here for backwards compatibility
     MIM = 111  # milligram
+    MCM = 116  # microgram
+    MIR = 74  # millilitre
+    MCR = 110  # microlitre
```
currycoder, line 215: While I understand the desire to make this cleaner/easier to bin later, I think possibly the opposite has happened.  The callers now all need to know that there's two buckets of mappings so implementation details have leaked further than they need to, and dropping these later on will be more complicated as you will need to unwind the new logic that supports both buckets.

I would suggest just having the `UnitMapping` bucket and keeping legacy/new values on there in tandem.  The legacy values can be marked in a very obvious way with comments.

hnryjmes: I think Brendan is right. There doesn't need to be a one-to-one correspondence between units and their numeric code. The relevant restriction is just that the keys are all unique, which it appears we have. I think we can have multiple units with the same numeric code, but let's write tests for this to demonstrate it working.

Here's a pretty picture that helped me get my head around it: https://en.wikipedia.org/wiki/Bijection,_injection_and_surjection#/media/File:Surjection.svg

hnryjmes: Or in simpler terms, it's not that we have legacy units, it's that we have legacy names for the same units

mail/enums.py
markj0hnst0n: Apologies, I'm just catching up on these comments as yesterday was quite busy.  @currycoder Your bullet points are basically the strategy I'd outlined to @hnryjmes when we spoke yesterday.

The backwards compatibility can be left in until removal.

The only thing I'd add to your points @currycoder  is that I'd planned on creating a migration in the LITE API which iterates through the relevant elements in the database converting the old elements into the new ones.  This would reduce the time the backwards compatible code needs to stay.

mail/enums.py
markj0hnst0n: @currycoder Removed the extra enum and just used the original UnitMapping.  While it proved how easy it was to remove the lean nature of the code that's left proves that you were absolutely right to call this out!

@hnryjmes Test data has been added to existing tests to confirm that the legacy code MIM works in the licence_payload_file

mail/tests/test_licence_to_edifact.py
markj0hnst0n: Removed

mail/serializers.py
markj0hnst0n: If a key is the same in both lists I don't think there there would be an error but a duplicate key in the same list would.  Each enum list is it's own entity governed by it's own internal rules but unfortunately I can't say for certain now as I've deleted this code!

currycoder (approved):
Thanks for your patience!

PR #277, opened 2024-09-16
Changes to allow deployment of the Dev branch to AWS using platform tools

james-francis-MT:
Configure application to run on DBT Platform while being backwards compatible with GOVPaas.


.copilot/phases/pre_build.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -e
+
+git_clone_base_url="https://codestar-connections.eu-west-2.amazonaws.com/git-http/730335529260/eu-west-2/192881c6-e3f2-41a9-9dcb-fcc87d8b90be/uktrade"
```
hnryjmes, line 5: this doesn't seem like the kind of thing that should be in a public git repo, unless I'm missing something

kevincarrogan (approved):

PR #278, opened 2024-09-27
Include number of unprocessed payloads in the error message

saruniitr:
## Change description

If the queue is completely stopped and the number of unprocessed payloads building up then the current error message cannot convey that so include the payload count in the error message.

There is also no need to emit message for each unprocessed payload because when we process them we batch together into a single message so the delay is same for all.

**Note:** We also prioritize sending SPIRE messages, as in before looking to batch unprocessed payloads we check if we are awaiting reply or any new messages from SPIRE that need sending. If they are all concluded then only we continue. This means during busy times then payload count can increase for a period and it is expected but should get cleared as the traffic subsides.


depsiatwal (approved):
lgtm

Tllew (approved):
Looks good to me

PR #281, opened 2024-11-04
DBT Platform Migration

acodeninja:
### Aim

Prepare the `lite-hmrc` codebase for migration to DBT platform from GOV PaaS.

[DPM-1283](https://uktrade.atlassian.net/browse/DPM-1283)

[DPM-1283]: https://uktrade.atlassian.net/browse/DPM-1283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal:
Tested on UAT deploying ok.

.copilot/phases/pre_build.sh
```diff
@@ -0,0 +1,18 @@
...
+git config --global credential.helper '!aws codecommit credential-helper $@'
+git config --global credential.UseHttpPath true
+
+cat <<EOF > ./.gitmodules
+[submodule "django_db_anonymiser"]
+	path = django_db_anonymiser
+	url = $git_clone_base_url/django-db-anonymiser.git
+EOF
+
+git submodule update --init --remote --recursive
```
currycoder, line 16: As per my comment about this same pattern in lite-api https://github.com/uktrade/lite-api/pull/2268#discussion_r1832774984

I feel like we might need to retain the fine grained control that we have for locking submodule versions in step with code releases.

depsiatwal: Also raised on same ticket for DBT Platform team to address this

conf/settings.py
```diff
@@ -230,6 +233,7 @@
         environment=env.str("SENTRY_ENVIRONMENT"),
         integrations=[DjangoIntegration()],
         send_default_pii=True,
+        traces_sample_rate=env.float("SENTRY_TRACES_SAMPLE_RATE", 1.0),
```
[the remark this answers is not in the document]
acodeninja: Yes we would, the current sentry instance cannot take 100% tracing samples in prod. We would look to have 100% on a dev environment and then somewhere between 2 and 5% on prod.

conf/settings.py
```diff
@@ -49,18 +49,21 @@
...
 
+if not IS_ENV_DBT_PLATFORM:
+    INSTALLED_APPS += [
+        "healthcheck",
+        "health_check.db",
+        "health_check.cache",
+        "health_check.storage",
+        "health_check.contrib.migrations",
+        "health_check.contrib.celery",
+        "health_check.contrib.celery_ping",
```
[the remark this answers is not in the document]
acodeninja: So in terms of the healthchecks, the `db`, `cache`, `storage` and `migrations` are probably ok, however the celery ones are a definite no go.

In DBT platform healthchecks should be scoped to checking only the service in question, rather than other services. As the web and celery processes run as separate services and each have their own healthchecks they should never check eachothers health.

.copilot/phases/pre_build.sh
acodeninja: See my comment on LTD-5627 this doesn't appear to be an issue.

currycoder: As per comments on the Jira, I believe it is an issue.  Given this one hasn't merged yet could the `--remote` flag be dropped?  This will maintain the pattern that we have been using for submodules up until now with gov-PAAS.

acodeninja: This is done.

currycoder (approved):

depsiatwal (approved):
lgtm

PR #284, opened 2024-11-26
LTD-5686-DBT-HMRC-Config

emileswarts:
Add additional configurations to allow this codebase to run as a scheduled task on DBT Platform.

### DIFF
--- a/.InstallPackages
+++ b/.InstallPackages
@@ -1,1 +1,0 @@
-postgresql-client-14
--- a/.InstallPackages
+++ b/.InstallPackages
@@ -1,0 +1,1 @@
+postgresql-client-14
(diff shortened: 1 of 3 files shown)


.InstallPackages
```diff
@@ -0,0 +1 @@
+postgresql-client-14
```
markj0hnst0n, line 1: I thought we agreed psql 13?

emileswarts: This is just a client to do the pg_dump. Found that it's unused and removed it. We use a buildpack to do this instead.

markj0hnst0n (approved):

depsiatwal (approved):
lgtm

PR #289, opened 2024-12-16
LTD-5736

markj0hnst0n:
add checking for error in rejected licence reply

### DIFF
--- a/mail/libraries/data_processors.py
+++ b/mail/libraries/data_processors.py
@@ -20,7 +20,7 @@ from mail.libraries.mailbox_service import find_mail_of
 from mail.models import LicenceData, Mail, UsageData
 from mail.serializers import LicenceDataMailSerializer, UpdateResponseSerializer, UsageDataMailSerializer
 
-# from mail.chief.licence_reply import LicenceReplyProcessor
+from mail.chief.licence_reply import LicenceReplyProcessor
 
 
 class EdifactFileError(Exception):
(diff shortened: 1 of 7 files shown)


mail/libraries/data_processors.py
```diff
@@ -53,6 +59,23 @@ def serialize_email_message(dto: EmailMessageDto) -> Mail or None:
...
+        processor = LicenceReplyProcessor.load_licence_reply_from_mail(_mail)
+
+        if processor._current_rejected:
+            rejected_transaction_errors = processor._current_rejected.errors
+            for error in rejected_transaction_errors:
+                if "Duplicate transaction reference" in error.text:
+                    run_number = LicenceData.objects.get(mail=_mail).hmrc_run_number
+                    raise EdifactFileError(
+                        f"Unable to process file due to the following error: {error.text}.  It was sent in run number {run_number}"
+                    )
```
saruniitr, line 72: This method is not the right place to raise as it will not update state correctly. If we raise here then we won't relay the reply message back to SPIRE. We also don't update the mail status which happens later.

or maybe that is the intention (to stop updating the status)?

markj0hnst0n: No I think this is probably a good point.  I think we want as much data as we can get as if the system is working until it fails.  I was thinking about putting it at the end of [check_and_notify_rejected_licences()](https://github.com/uktrade/lite-hmrc/blob/f27182031be860d0cea01ce3f73ee485aa78d4ba/mail/libraries/routing_controller.py#L222)

saruniitr: Yeah I was going to suggest the same because by that point mail is completely processed and status is updated. However note that it also means task is going to start processing next mail (if available). Only thing that stops the queue is if there is any reply pending. So if we see this error we need a way to suspend the queue from processing new messages (after the status is updated).

PR #295, opened 2024-12-22
Move mailboxes code to separate app

kevincarrogan:
At the moment nearly all of the `lite-hmrc` code resides in a single Django application called `mail`. This makes it quite overwhelming to understand where to look for code when debugging and diagnosing problems.

The main purpose of this PR is to split out one aspect of the `mail` application which is the code related to mail servers and mailboxes.

As part of this there are a few bonus extras as well:

  - Adding integration tests for these new apps
  - Allowing mail servers to be fully defined in config (this makes testing easier)
  - Removal of unused code (mock_hmrc and ICMS related code)
  - Removal of Django Choices code as we now have it available directly from Django
  - Simplified a lot of email parsing using the Python `email` standard library
  - Rework the pulling of messages from a mailbox to, hopefully, make it more understandable
  - Add a _lot_ of tests to cover all of above changes


mailboxes/migrations/0001_initial.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [("mail", "0023_remove_mailreadstatus_mailbox_delete_mailboxconfig_and_more")]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
```
currycoder, line 18: Basically take control of the MailboxConfig model from the `mailboxes` app without dropping the table?

mail/management/commands/mark_mails_as_read.py
```diff
@@ -4,6 +4,7 @@
 from mail import enums, models
 from mail.libraries.mailbox_service import get_message_id
```
currycoder: shouldn't this be removed?

currycoder: This makes me wonder if there should be a test for this command somewhere

mailboxes/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    msg_num = listing_msg.split()[0]
+
+    # retrieves the header information
+    # 0 indicates the number of lines of message to be retrieved after the header
+    msg_header = pop3_connection.top(msg_num, 0)
+
+    spire_from_address = settings.SPIRE_FROM_ADDRESS.encode("utf-8")
+    hmrc_dit_reply_address = settings.HMRC_TO_DIT_REPLY_ADDRESS.encode("utf-8")
+
+    if spire_from_address not in msg_header[1] and hmrc_dit_reply_address not in msg_header[1]:
```
currycoder: I know this is lifted/shifted code but `msg_header[1]` crops up a lot and it would be more readable to have that as a meaningfully named variable

mailboxes/utils.py
```diff
@@ -3,9 +3,17 @@
 from email.parser import BytesHeaderParser
 from email.utils import parseaddr
 
+from django.conf import settings
+
 logger = logging.getLogger(__name__)
 
 
+def get_message_number(listing_message):
```
currycoder: I feel like some docstrings on these utils would be a help for my understanding.

mailboxes/utils.py
```diff
@@ -35,12 +34,14 @@ def get_message_id(msg_header):
...
-            mailbox=mailbox_config, status__in=[MailReadStatuses.READ, MailReadStatuses.UNPROCESSABLE]
-        )
-    ]
+    mail_read_statuses = mailbox_config.mailreadstatus_set.filter(
+        status__in=[
+            MailReadStatuses.READ,
+            MailReadStatuses.UNPROCESSABLE,
+        ]
+    )
+    read_message_ids = set(mail_read_statuses.values_list("message_id", flat=True))
```
currycoder, line 50: This set of message ids gets larger and larger, no?

mail/libraries/mailbox_service.py
```diff
@@ -52,60 +52,63 @@ def get_message_iterator(pop3_connection: POP3_SSL, username: str) -> Iterator[T
...
+            """
+            logger.info(
+                "Marking message_id %s with message_num %s from %s as %s",
+                message_id,
+                message_num,
+                read_status.mailbox,
+                status,
+            )
+            read_status.status = status
+            read_status.save()
```
currycoder: This really is something

mail/libraries/mailbox_service.py
```diff
@@ -9,53 +9,19 @@
...
 from mailboxes.enums import MailReadStatuses
 from mailboxes.models import MailboxConfig
-from mailboxes.utils import (
-    get_message_header,
-    get_message_id,
-    get_message_number,
-    get_read_messages,
-    is_from_valid_sender,
-)
+from mailboxes.utils import get_unread_message_ids
```
currycoder: Lovely

mailboxes/utils.py
```diff
@@ -77,3 +83,69 @@ def is_from_valid_sender(msg_header, valid_addresses):
...
+    def __call__(self, status):
+        logger.info(
+            "Marking message_id %s with message_num %s from %r as %s",
+            self.message_id,
+            self.message_num,
+            self.read_status.mailbox,
+            status,
+        )
+        self.read_status.status = status
+        self.read_status.save()
```
currycoder: Can't this just be a model method..?

mailboxes/utils.py
```diff
@@ -51,101 +51,125 @@ def get_read_messages(mailbox_config):
...
+    _, mails, _ = pop3_connection.list()
+    mails = mails[-max_limit:]
+    for mail in mails:
+        message_number = get_message_number(mail)
+        message = MailboxMessage(pop3_connection, mailbox_config, message_number)
+        yield message
+
+
+def is_read(message, read_messages):
+    return message.message_id in read_messages
```
currycoder, line 115: Wondering if `MailboxMessage` can have the read status passed in to it on creation so that this could be a method on there instead?

mailboxes/utils.py
```diff
@@ -164,10 +164,21 @@ def get_message_iterator(server: MailServer) -> Iterator[Tuple[EmailMessageDto,
...
                 continue
 
-            read_status, _ = mailbox_config.mail_read_statuses.get_or_create(
+            logger.debug(
+                "About to create or update mail_read_status for %s (%s)", message.message_id, message.message_number
+            )
+            read_status, created = mailbox_config.mail_read_statuses.update_or_create(
                 message_id=message.message_id,
                 message_num=message.message_number,
-                mail_data=mail_data,
```
currycoder: How can the email message data be mutable..?

mailboxes/migrations/0001_initial.py
kevincarrogan: Yep. That’s exactly what this all does.

mail/management/commands/mark_mails_as_read.py
kevincarrogan: It’s already gone.

mailboxes/utils.py
kevincarrogan: This should be changed already.

mailboxes/utils.py
kevincarrogan: It does. This keeps the implementation mostly the same as it was before but switching to a set should mean a slightly faster lookup instead of gilding a big list and checking existence that way.

mailboxes/utils.py
kevincarrogan: Probably. But this is keeping the output of this function the same so I didn’t have to change it at the point where it’s used.

mailboxes/utils.py
kevincarrogan: I think this would end up being mostly the same in the end anyway, it’s just slightly shifting whether it’s a function or a method. You would still end up doing the same calls really.

mailboxes/utils.py
kevincarrogan: It was to do with headers that came back from the mailbox. Technically the email is identical but there are headers that change when you read the message multiple times, something like a request id or an updating timestamp. 

In this case it was to stop the message from being duplicated if the email gets read again as the data isn’t really changing.

kevincarrogan: But I chose to do the update so it preserves the most recent headers in the case you did read the file data multiple times.

mail/libraries/mailbox_service.py
kevincarrogan: I switched this to an object as I did find some weird capturing anomalies when this closure was called later in some of my tests where the `read_status` object was actually changing for all of these closures instead of each of them having their own read_status object. 

But using the __call_ method at least meant I didn’t need to change the api for the calling code.

mailboxes/utils.py
currycoder: So just to fully check my understanding, we get all of the IDs for the messages read since the start of time in the app?  Any ideas how large this is right now?

mailboxes/utils.py
currycoder: Got it, could I just suggest a comment to make this obvious for the next dev?

mailboxes/utils.py
kevincarrogan:
```
>>> from mail.models import MailReadStatus
>>> MailReadStatus.objects.count()
38029
```

mailboxes/utils.py
currycoder: Discussed on slack.  The reason is that this is like a kind of callback which is passed to callers outside of this app in step with each message DTO.  So instead of changing the contract with those other callers, the contract is maintained by doing it like this.  We may decide to change that contract later but for now this keeps the refactor more pure.

mailboxes/utils.py
currycoder: Thanks.  Seems fine for now then.

currycoder (approved):

depsiatwal (approved):
having been thought this in some detail makes sense to be and looks good.

PR #296, opened 2024-12-23
Fix communication between LITE API and LITE HMRC

markj0hnst0n:
On DBT Platform there is a URL mismatch between the HMRC and API apps which causes a hawk auth failure on the HMRC side.  There is an unusual piece of code which replaces http with https which causes the Message Authentication Code to change which appears to have been a hack to get the auth to work correctly on gov paas.

Once this is removed the systems communicate normally but this change must be backwards compatible with Gov Paas.

Deployed on dev data environment for testing and some short testing notes are included on the ticket

[LTD-5751](https://uktrade.atlassian.net/browse/LTD-5751)

[LTD-5751]: https://uktrade.atlassian.net/browse/LTD-5751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/conf/authentication.py
+++ b/conf/authentication.py
@@ -49,7 +49,7 @@ def _authenticate(request):
 
     # TODO: remove this when migration to DBT platform is complete as this hack is only required on Gov Paas
     if is_env_gov_paas():
-        convert_url(url)
+        convert_gov_paas_url(url)
 
     if hawk_authentication_enabled():
         return Receiver(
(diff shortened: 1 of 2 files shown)


kevincarrogan (approved):

depsiatwal (approved):
lgtm

PR #298, opened 2024-12-23
Split out logic for checking and routing emails

kevincarrogan:


conf/settings.py
```diff
@@ -47,6 +47,7 @@
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "mail",
```
depsiatwal, line 49: will the mail app  be removed eventually

conf/settings.py
```diff
@@ -138,6 +129,35 @@
 SPIRE_ADDRESS = env("SPIRE_ADDRESS", default="test-spire-address@example.com")  # /PS-IGNORE
 HMRC_ADDRESS = env("HMRC_ADDRESS", default="test-hmrc-address@example.com")  # /PS-IGNORE
 
+AZURE_AUTH_CLIENT_ID = env.str("AZURE_AUTH_CLIENT_ID")
+AZURE_AUTH_CLIENT_SECRET = env.str("AZURE_AUTH_CLIENT_SECRET")
+AZURE_AUTH_TENANT_ID = env.str("AZURE_AUTH_TENANT_ID")
+
+MAIL_SERVERS = {
```
depsiatwal, line 136: nice this makes our settings much easier to understand

depsiatwal (commented):
A lot of changes here and  all look sensible to me. 

I don't understand what all the code is doing but I get the idea. Remove all the mailbox logic and decouple it everywhere we sending emails. 

I'm just wondering how we test these changes are safe is there a plan to deploy your HMRC changes is a phased approach ?

i.e Mail box decoupling followed by edifact changes once we know the former has bedded in.

PR #299, opened 2024-12-27
Only add APM app if env setting is set

kevincarrogan:
Remove adding APM to installed apps when there is no URL set.

This was causing lots of logging errors.


saruniitr (approved):

depsiatwal (approved):
lgtm

PR #300, opened 2024-12-27
End to end tests

kevincarrogan:
Add end to end tests to verify various full flows of information from SPIRE, HMRC and LITE.

I made quite a couple of design decisions whilst doing this.

_Always use TLS for SMTP_
This is to minimise the code paths available and make sure that the service is secure by default.

_Switched from Mailhog to Mailpit_
  - Mailhog is currently unmaintained
  - Mailpit allowed me to use TLS in tests

_Used pytest style tests_
I initially started writing the tests using unittest but the class based approach using the existing LITEDataClient did a bunch of setup that I started to undo and this pushed me to using pytest and using fixtures instead to setup tests with just what they need.


conf/mail_servers/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine AS certs
+
+ARG CN
+RUN apk --update add openssl
+RUN openssl req -batch -x509 -newkey rsa:4096 -nodes -keyout key.pem -out cert.pem -sha256 -days 3650 -subj "/CN=${CN}"
+
+FROM axllent/mailpit
+RUN mkdir /certs
+COPY --from=certs *.pem /certs
```
hnryjmes, line 9: trying to follow along... so this `Dockerfile` set up is used instead of putting 

```
services:
  mailpit:
    image: axllent/mailpit
    ...
```

 in `docker-compose.yml`? and the reason for doing that is to do with TLS certs?

kevincarrogan: Correct.

mail/libraries/builders.py
```diff
@@ -268,8 +267,8 @@ def build_email_message(email_message_dto: EmailMessageDto) -> MIMEMultipart:
         "Content-Disposition",
         f'attachment; filename="{email_message_dto.attachment[0]}"',
     )
-    payload.add_header("Content-Transfer-Encoding", "7bit")
     payload.add_header("name", email_message_dto.subject)
+    payload.replace_header("Content-Transfer-Encoding", "7bit")
```
saruniitr, line 271: Without this header I think there were problems getting the file recognized on the HMRC systems.
Looking at the documentation, `replace_header()` raises error if the header is not present so assuming the header is always included then this change is good.

docker-compose.yml
```diff
@@ -1,15 +1,10 @@
-version: "3.8"
-
 services:
   lite-hmrc-postgres:
+    image: "postgres:12"
```
depsiatwal, line 3: why do we need 12 and 13 ?
for DBT right ? think I had a draft PR that removes it so would probably be easier my PR to  base of this now.

mail/tests/test_end_to_end.py
```diff
@@ -1,238 +1,520 @@
...
+        return_value=spire_to_dit_mailserver,
+    )
+
+
+def normalise_line_endings(string):
+    return string.replace("\r", "").strip()
+
+
+@pytest.fixture()
+def get_smtp_message_count(outbox_mailserver_api_url):
```
depsiatwal, line 142: quite a few fixtures here is it more readable if we placed in conftest ?

mail/tests/test_end_to_end.py
```diff
@@ -1,238 +1,520 @@
...
+    mail.refresh_from_db()
+    assert mail.status == ReceptionStatusEnum.REPLY_SENT
+    assert normalise_line_endings(mail.response_data) == normalise_line_endings(licence_reply_file_body.decode("ascii"))
+    assert mail.response_filename == licence_reply_file_name
+    assert mail.response_subject == licence_reply_file_name
+
+    assert get_smtp_message_count() == 0
+
+
+def test_receive_lite_usage_data_from_hmrc_e2e(
```
depsiatwal, line 278: infact a lot of tests here I know they're classed as e2e but wondering if it's cleaner to break down into logical order with relevant filenames .

docker-compose.yml
kevincarrogan: This file is for local development and it’s set to 12 as that’s what we currently run in production. 

Before this change it wasn’t actually pinned to any specific version of Postgres so was potentially pulling much later version than we run in production.

mail/libraries/builders.py
kevincarrogan: Yes. The header is present as we are building it all above and so it would be failing in our tests with said error if it wasn’t. 

in my testing what was happening was that both headers were being sent and this was messing things up in some mail clients and mailpit.

saruniitr (approved):
all changes looks good to me

mail/tests/test_end_to_end.py
kevincarrogan: I've found that putting everything into conftest too early actually makes things more of a mess.

The fixtures are only required for this particular set of tests and as such I have put them in this file, if they are required for more tests then it would be suitable to put them in conftest.

mail/tests/test_end_to_end.py
kevincarrogan: Can you describe what you mean?

depsiatwal: actually ignore that comment just viewed the whole file again and it's not that bad.

depsiatwal (approved):
LGTM

hnryjmes (approved):
nice job

PR #301, opened 2025-01-02
Ltd 5761 edifact processing

kevincarrogan:


edifact/grammars/usage_data.lark
```diff
@@ -0,0 +1,113 @@
+_m{x}: _BACKSLASH x
```
depsiatwal, line 1: nice like it

edifact/management/commands/validate_usage_data_parser.py
```diff
@@ -0,0 +1,64 @@
...
+                continue
+
+            try:
+                transformed_edi_data = Edifact().transform(tree)
+            except LarkError:
+                logger.warning("Failed to transform tree into edifact %s", usage_data, exc_info=True)
+                self.stdout.write(self.style.ERROR(f"Failed to transform tree for {usage_data}"))
+                continue
+
+            if edi_data != transformed_edi_data:
```
depsiatwal, line 37: quite difficult to follow probably worth considering a pattern here

PR #303, opened 2025-01-10
Update codestar connection step

james-francis-MT:
Update the pre build step to match the standard way of pulling submodules


markj0hnst0n (approved):

depsiatwal:
This is not a prerequisite for go-live hence  we should be freezing such changes now hence this will be picked up after go-live

PR #306, opened 2025-02-06
LTD-5853 unidecode before textwrapping

Tllew:
**Issue:**
* hmrc requires a length of 35 per address line
* lite-hmrc achieved this using textwrap
* unidecode appended to the character count and we managed to go over

**Things to note:**
I had initially added the unidecode to edifact_validators so that the validator could catch the error and went back and forth with this idea for a little while, in theory because it's decoded now in the sanitizer (lite_to_edifact_converter) it will never hit the sanitizer with special characters present.
I have also added some coverage tests in test_builder so that we cant remove this and accidentally have things pass. 
if we notice any other edge cases we could update the new payload file to include them

[LTD-5853](https://uktrade.atlassian.net/browse/LTD-5853)

[LTD-5853]: https://uktrade.atlassian.net/browse/LTD-5853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mail/libraries/lite_to_edifact_converter.py
+++ b/mail/libraries/lite_to_edifact_converter.py
@@ -5,6 +5,7 @@ import textwrap
 from typing import TYPE_CHECKING, Dict, Iterable, Optional
 
 from django.utils import timezone
+from unidecode import unidecode
 
 from mail.enums import (
     LITE_HMRC_LICENCE_TYPE_MAPPING,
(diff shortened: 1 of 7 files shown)


saruniitr (approved):
looks good to me.

mail/libraries/lite_to_edifact_converter.py
```diff
@@ -247,7 +248,7 @@ def sanitize_foreign_trader_address(trader):
     addr_line = addr_line.replace("\n", " ").replace("\r", " ")
     # replace special characters
     addr_line = addr_line.replace("#", "")
-    addr_lines = textwrap.wrap(addr_line, width=FOREIGN_TRADER_ADDR_LINE_MAX_LEN, break_long_words=False)
+    addr_lines = textwrap.wrap(unidecode(addr_line), width=FOREIGN_TRADER_ADDR_LINE_MAX_LEN, break_long_words=False)
```
depsiatwal, line 251: so for my understanding it breaks because we do validation (all is less then 35) Then we do unicode before sending and it goes over 35 ?

Also is this validation only and doesn't impact the save ?

Tllew: no this will be unidecoding the address line, I suppose in theory you could then reencode the outputted textwrap but it seems like more work to avoid something that happens in a later step anyway

depsiatwal (approved):
lgtm

PR #309, opened 2025-03-28
REPLACE payloads causing failures

kevincarrogan:
The `REPLACE` action has been used for payloads that were sent but then subsequently needed re-sending because of special characters.

The problem with using `REPLACE` is that this ended up breaking an implicit contract where a part of code expected there to be only one `LicencePayload` with a reference.

This fix stops the expectation of there only being a single `LicencePayload` object for a given reference.


mail/libraries/lite_to_edifact_converter.py
```diff
@@ -28,10 +28,27 @@
...
 class EdifactValidationError(Exception):
     pass
 
 
+class PreviousPayloadError(Exception):
+    pass
+
+
+def get_previous_licence_payload(old_reference):
+    previous_payloads = LicencePayload.objects.filter(reference=old_reference)
```
saruniitr, line 43: just wondering if it is safer to also filter by `is_processed=True` and previous means should've been processed already.

saruniitr (approved):
looks good to me

Tllew (approved):

PR #310, opened 2025-04-28
Update licence data serializers

hnryjmes:
This change preserves the exact logic for required fields etc that we had before, it just refactors the code so instead of if-statements we have separate serializers and a mapping between specific licence types and specific serializers.

The tests have also been updated so we are testing specific serializers instead of that generic serializer.

Also added a test for the new NotImplementedError.

[LTD-6174](https://uktrade.atlassian.net/browse/LTD-6174)

[LTD-6174]: https://uktrade.atlassian.net/browse/LTD-6174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mail/tests/test_serializers.py
+++ b/mail/tests/test_serializers.py
@@ -6,12 +6,16 @@ from rest_framework.exceptions import ErrorDetail
 
 from mail import icms_serializers
 from mail.enums import ChiefSystemEnum, LicenceTypeEnum, UnitMapping
-from mail.serializers import LiteLicenceDataSerializer, LiteStandardLicenceDataSerializer
+from mail.serializers import (
+    LiteGenericLicenceDataSerializer,
+    LiteOpenIndividualExportLicenceDataSerializer,
+    LiteStandardIndividualExportLicenceDataSerializer,
+)
 
 
 class LiteLicenceDataSerializerTestCase(TestCase):
     def test_no_data(self):
-        serializer = LiteLicenceDataSerializer(data={})
+        serializer = LiteGenericLicenceDataSerializer(data={})
 
         self.assertFalse(serializer.is_valid())
         expected_errors = {
@@ -33,7 +37,7 @@ class LiteLicenceDataSerializerTestCase(TestCase):
             "start_date": "1999-12-31",
             "type": "baz",
         }
-        serializer = LiteLicenceDataSerializer(data=data)
+        serializer = LiteGenericLicenceDataSerializer(data=data)
 
         self.assertFalse(serializer.is_valid())
         expected_errors = {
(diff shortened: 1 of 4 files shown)


mail/views.py
```diff
@@ -76,17 +78,13 @@ def post(self, request):
...
-            return serializers[app_type]
-
-        return LiteLicenceDataSerializer
+        app_type_to_serializer = {
+            str(LicenceTypeEnum.SIEL): LiteStandardIndividualExportLicenceDataSerializer,
+            str(LicenceTypeEnum.OIEL): LiteOpenIndividualExportLicenceDataSerializer,
+        }
+        if app_type in app_type_to_serializer.keys():
+            serializer = app_type_to_serializer[app_type]
+            return serializer
```
kevincarrogan, line 87: What happens if we don't find the serializer for our licence type?

hnryjmes: it should return None and then an error will be thrown when the serializer can't be used as a serializer

I thought about having an explicit error though. what do you think? 

I originally had this work so it fell through and used the generic licence data serializer directly but I think that's a bad idea and we shouldn't encourage it.

kevincarrogan: I think an explicit error would be nice instead of just letting something else fail when it expects a serializer but then receives `None`.

kevincarrogan (approved):

mail/views.py
hnryjmes: ok, I'll look into how we do that

hnryjmes: updated

mail/tests/test_serializers.py
```diff
@@ -208,13 +161,74 @@ def test_goods_valid_choice_for_unit(self):
...
 
 
+class LiteOpenIndividualExportLicenceDataSerializerTestCase(TestCase):
+    def test_required_fields_for_oiel_type(self):
+        data = {
+            "action": "insert",
+            "end_date": "1999-12-31",
+            "id": "foo",
+            "reference": "bar",
+            "start_date": "1999-12-31",
```
saruniitr, line 178: Start and end dates can be same for OIELs?
Could we use `freeze_time` and update the date to current dates here and in below tests?

hnryjmes: the serializer doesn't do any smart checking of these date strings, it just checks if it's any ISO date. are you saying we should make the serializer check if the dates make business sense?

saruniitr: I am sure there will be validation where these dates are entered by the user on the form to check if they are valid dates, not in the past etc.
This serializer is ensuring some of the fields are marked as required or not based on the type so it wouldn't hurt validating the dates as well but considering this is only refactoring we may not want to make any new functional changes may be it can be a follow-up.

hnryjmes: yeah I'm open to putting more validation in these serializers, and make them more useful, but it could maybe be a task we do later. I only think we should have redundant data validation like this if it really serves a useful purpose.

saruniitr: forgot to mention earlier, for this PR please update the dates to reflect correct validity dates based on the types so the test data makes sense.

saruniitr (approved):
With the test data adjusted it looks good to me.

PR #312, opened 2025-05-12
Update lite_to_edifact_converter

hnryjmes:
The test file `test_licence_to_edifact.py` has been updated with a new test `test_single_oiel` which shows that, given a LicencePayload for an OIEL is created, that the `licences_to_edifact` function in `lite_to_edifact_converter.py` handles the LicencePayload correctly and the expected edifact data is produced.

The test file `open_licence_payload_file` is a JSON file based on the existing `licence_payload_file` in the same place for a SIEL. It's a work in progress as we have not finalised the JSON payload from lite-api to send an OIEL to lite-hmrc but this is a first approximation. 

Changes from the SIEL payload include:
- No `end_user` field
- No `goods` list field
- A new `countries` list field with the country data.

This was based on existing OIEL payload data used for OIELs from SPIRE.

None of the existing code for handling SIELs in LITE-HMRC has been changed and everything should work exactly the same even with these changes to the tests.

We can extend the test coverage for OIELs in the `test_licence_to_edifact.py` file once we know a bit more about what we want to test.

EDIT: Additionally `controlled_by` is now set to `"O"` which is actually a requirement for the edifact file that was missed before.

### DIFF
--- a/mail/libraries/edifact_validator.py
+++ b/mail/libraries/edifact_validator.py
@@ -231,7 +231,7 @@ def validate_licence_product_line(record):
     # Use chief type rather than indexing into tokens
     ld = chieftypes.LicenceDataLine(*tokens)
 
-    if ld.goods_description == "Open Licence goods - see actual licence for information":
+    if ld.controlled_by == "O":
         # open licence goods, skip further checks
         return errors
 
--- a/mail/libraries/edifact_validator.py
+++ b/mail/libraries/edifact_validator.py
@@ -231,6 +231,10 @@ def validate_licence_product_line(record):
     # Use chief type rather than indexing into tokens
     ld = chieftypes.LicenceDataLine(*tokens)
 
+    if ld.goods_description == "Open Licence goods - see actual licence for information":
+        # open licence goods, skip further checks
+        return errors
+
     if settings.CHIEF_SOURCE_SYSTEM == "SPIRE":
         if not ld.goods_description:
             errors.append({record_type: "Product description cannot be empty"})
--- a/mail/tests/files/open_licence_payload_file
+++ b/mail/tests/files/open_licence_payload_file
@@ -1,0 +1,37 @@
+{
+  "licence": {
+    "id": "3de27a3e-defa-4d59-8223-947d2dc4bb00",
+    "reference": "GBOIEL/2026/0000001/P",
+    "type": "oiel",
+    "action": "insert",
+    "start_date": "2026-06-02",
+    "end_date": "2031-06-02",
+    "organisation": {
+      "name": "Organisation",
+      "id": "10a21d56-9e9d-333d-77c5-479bb3de7ac9",
+      "eori_number": "GB123456789000",
+      "address": {
+        "line_1": "might",
+        "line_2": "248 James Key Apt. 515",
+        "line_3": "Apt. 942",
+        "line_4": "West Ashleyton",
+        "line_5": "Farnborough",
+        "postcode": "GU40 2LX",
+        "country": {
+          "id": "GB",
+          "name": "United Kingdom"
+        }
+      }
+    },
+    "countries": [
+      {
+        "id": "CA",
+        "name": "Canada"
+      },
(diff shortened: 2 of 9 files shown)


mail/tests/libraries/client.py
```diff
@@ -37,11 +34,8 @@ def setUp(self):
 
         self.licence_payload_json = json.loads(read_file("mail/tests/files/licence_payload_file", encoding="utf-8"))
 
-        self.single_siel_licence_payload = LicencePayload.objects.create(
-            lite_id=self.licence_payload_json["licence"]["id"],
-            reference=self.licence_payload_json["licence"]["reference"],
-            data=self.licence_payload_json["licence"],
-            action=LicenceActionEnum.INSERT,
```
hnryjmes, line 44: this code for LicencePayload object creation being in the shared test client means that db data for tests aren't properly isolated if they use this client code. 

it's better to move it to a specific `setUp()` function in the test classes for SIELs that need this LicencePayload object created for the tests to function.

previously we only considered SIELs so it wasn't a problem before.

mail/libraries/edifact_validator.py
```diff
@@ -231,6 +231,10 @@ def validate_licence_product_line(record):
     # Use chief type rather than indexing into tokens
     ld = chieftypes.LicenceDataLine(*tokens)
 
+    if ld.goods_description == "Open Licence goods - see actual licence for information":
+        # open licence goods, skip further checks
+        return errors
+
```
hnryjmes, line 237: ideally we should refactor this whole function to not make it specific to SIELs, but for now just doing this hack

mail/tests/files/open_licence_payload_file
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "countries": [
+      {
+        "id": "CA",
+        "name": "Canada"
+      },
+      {
+        "id": "US",
+        "name": "United States"
+      }
```
hnryjmes, line 34: These countries are taken from the example OIEL edifact data provided in the ideas document

mail/tests/test_licence_to_edifact.py
```diff
@@ -358,7 +366,40 @@ def test_trader_address_sanitize(
...
+        super().setUp()
+        self.single_oiel_licence_payload = LicencePayload.objects.create(
+            lite_id=self.open_licence_payload_json["licence"]["id"],
+            reference=self.open_licence_payload_json["licence"]["reference"],
+            data=self.open_licence_payload_json["licence"],
+            action=LicenceActionEnum.INSERT,
+        )
+
+    def test_single_oiel(self):
+        licences = LicencePayload.objects.filter(is_processed=False)
```
saruniitr, line 380: Could you add another test where there are multiple OIELs?
If there are multiple payloads then we batch and include multiple licences in the same file so I think we need to cover that case as well.

mail/libraries/edifact_validator.py
saruniitr: this works but as per the comment https://github.com/uktrade/lite-hmrc/blob/master/mail/libraries/chieftypes.py#L182,
`controlled_by` is going to be `O` for OIELs, is it better to use that here?

mail/tests/test_licence_to_edifact.py
saruniitr: Maybe this will come in future, some of the countries can be allowed/prohibited so we need to add tests to cover those cases as well.

Tllew (approved):
Looks good

mail/libraries/edifact_validator.py
hnryjmes: I checked and you're right, it should be "O" (`O`) but this is not being set properly. The SPIRE OIELs have this as well. Will update and try to use that in my test

hnryjmes: updated

saruniitr (approved):
looks good to me.
Is it that the multiple payload test will be done later?

